PR #5, opened 2022-10-17
Fix SDK name for macOS

mtj0928:
The SDK name for macOS is not correct.
As a result, scipio cannot make xcframewrok for macOS.
I fixed the SDK name in this PR.

### DIFF
--- a/Sources/ScipioKit/Compiler.swift
+++ b/Sources/ScipioKit/Compiler.swift
@@ -239,7 +239,6 @@ struct Compiler<E: Executor> {
                 ("scheme", context.target.name),
                 ("archivePath", xcArchivePath.pathString),
                 ("destination", context.sdk.destination),
-                ("sdk", context.sdk.name),
             ].map(Pair.init(key:value:))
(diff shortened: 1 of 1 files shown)


giginet (approved):
Thanks!
I didn't care about macOS at all 😅 

By the way, these implementations are inspired by swift-create-xcframework.
https://github.com/unsignedapps/swift-create-xcframework/blob/main/Sources/CreateXCFramework/Platforms.swift#L23-L31

So it also would be broken...

mtj0928:
@giginet Thank you for your comment!
I found my change is not correct, because it makes some unexpected side effects.

Also, I found a new resolution.
`archive` command doesn't require `sdk`, so how about removing the options from `ArchiveCommand`.
If it's OK, I'll fix it, so let me know.
(FYI: swift-create-xcframework doesn't specify `sdk` for the command.)

mtj0928:
Sure!!

giginet (approved):
Thank you!

PR #7, opened 2022-10-20
Inroduce SwiftLint

giginet:
Introduce SwiftLint

I tried to integrate SwiftLint with SwiftPM build tools plugin. 
However, as a result, I don't include SwiftPM in this package in some reasons.


rsahara (approved):
LGTM

PR #15, opened 2022-12-02
Refactor Compiler

giginet:
For next feature development, first, refactor `Compiler`.

I added `FrameworkProducer` to produce XCFramework.
This producer uses `Compiler` and `CacheSystem`.


Sources/ScipioKit/Producer/Compiler.swift
```diff
@@ -0,0 +1,109 @@
...
+            return DebugSymbol(dSYMPath: dsymPath,
+                               target: target,
+                               sdk: sdk,
+                               buildConfiguration: buildConfiguration)
+        }
+        // You can use AsyncStream
+        var symbolMapPaths: [URL] = []
+        for dSYMs in debugSymbols {
+            let maps = try await self.extractor.dump(dwarfPath: dSYMs.dwarfPath)
+            let paths = maps.values.map { uuid in
```
freddi-kit, line 87: I prefer to use more specific name than map

```suggestion
            let dumpedDSYMsMaps = try await self.extractor.dump(dwarfPath: dSYMs.dwarfPath)
            let paths = dumpedDSYMsMaps.values.map { uuid in
```

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -0,0 +1,136 @@
...
+                                      buildOptions: buildOptions,
+                                      outputDirectory: outputDir,
+                                      storage: cacheStorage)
+        let compiler = Compiler(rootPackage: rootPackage, buildOptions: buildOptions)
+        try await compiler.clean()
+
+        let products = productsToBuild(for: mode)
+        for product in products {
+            if product.isBinaryTarget {
+                logger.error("🚧 \(product.target) is binaryTarget. Currently, binaryTarget is not supported. Skip it.")
```
freddi-kit, line 48: I think warning is better

Sources/ScipioKit/Runner.swift
```diff
@@ -62,17 +62,6 @@ public struct Runner {
             case disabled
             case project
             case storage(any CacheStorage)
-
```
freddi-kit, line 65: 👍

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -0,0 +1,136 @@
...
+                return
+            } else {
+                logger.warning("⚠️ Existing \(frameworkName) is outdated.", metadata: .color(.yellow))
+                logger.info("💥 Delete \(frameworkName)", metadata: .color(.red))
+                try fileSystem.removeFileTree(at: outputPath)
+                let restored = await cacheSystem.restoreCacheIfPossible(product: product)
+                needToBuild = !restored
+                if restored {
+                    logger.info("✅ Restore \(frameworkName) from cache storage", metadata: .color(.green))
+                }
```
freddi-kit, line 96: Why can we restore cache even the cache is invalid?

Sources/ScipioKit/Producer/FrameworkProducer.swift
giginet:
```suggestion
                logger.warning("🚧 \(product.target) is binaryTarget. Currently, binaryTarget is not supported. Skip it.")
```

Sources/ScipioKit/Producer/FrameworkProducer.swift
giginet: The invalid caches exist on an output dir ( e.g. `$PROJECT_DIR/XCFrameworks`)

On the other hand, Scipio tries to restore from other cache dirs (e.g. `~/Library/Caches`).

If XCFrameworks in output dirs are invalid, then they will be restored from cache dir.
So I think this implementation is correct.

freddi-kit (approved):

PR #18, opened 2022-12-05
Binary target extractor

### LINKED ISSUE #2: Does this support binary frameworks?
I am trying the following dependency
https://github.com/PSPDFKit/PSPDFKit-SP

And getting the below error
```
scipio prepare -v
evaluating manifest for 'dependencies' v. unknown
evaluating manifest for 'pspdfkit-sp' v. 11.5.2
evaluating manifest for 'dependencies' v. unknown
🔁 Resolving Dependencies...
/usr/bin/xcrun swift package resolve
Error: InternalError(description: "Internal error. Please file a bug at https://bugs.swift.org with this info. binary not supported")
```

The Package.swift of PSPDFKit [here](https://github.com/PSPDFKit/PSPDFKit-SP/blob/master/Package.swift)

Running `swift package resolve` on its own does not return any errors (or any output)

Is that something that can be worked around?

My dependency Package.swift [here](https://gist.github.com/alexandre-g/ce5067b3cf5b5b66b15a5d713f6485e6)

giginet:
This PR proceeded from #15

closes #2

I introduced `BinaryExtractor`.
This just copies XCFrameworks from `binaryTarget` in packages.


Tests/ScipioKitTests/RunnerTests.swift
```diff
@@ -53,8 +58,6 @@ final class PrepareTests: XCTestCase {
     }
 
     func testCacheIsValid() async throws {
-        try fileManager.createDirectory(at: frameworkOutputDir, withIntermediateDirectories: true)
-
```
giuk-jung, line 57: Why this line was deleted?

giuk-jung: Ah, I see. thank you

freddi-kit (approved):

PR #22, opened 2022-12-26
Support Resource bundle package

### LINKED ISSUE #11: Support resource bundle

giginet:
closes #11

Support resource bundle.

1. If a package contains resources, Collect all resources. it uses implementation inside SwiftPM.
2. Generate `{package_name}-Resources` target in xcodeproj. it is a resource bundle target.
3. Add the resource bundle target as a dependency to the framework target
4. Inject `resource-accessor.swift` into the framework target. It allows frameworks to use `Bundle.module`.
5. Build the framework, then the resource bundle is embedded into the framework.

Screen Shot 2022-12-26 at 18 02 43


.swiftlint.yml
```diff
@@ -9,14 +9,14 @@ disabled_rules:
   - force_try
   - todo
 line_length: 150
-file_length: 550
+file_length: 600
```
giginet, line 12: In the current implementation, ProjectGenerator is too huge.
It's better to refactor this. however, it makes this PR to be complex.

So I change the swiftlint settings looser.

Sources/ScipioKit/ProjectGenerator/ResourceCollector.swift
```diff
@@ -0,0 +1,53 @@
...
+            observabilityScope: observabilitySystem.topScope)
+    }
+
+    private let package: ResolvedPackage
+    private let target: ResolvedTarget
+    private let fileSystem: FileSystem
+
+    func collect() throws -> [Resource] {
+        let builder = try makeTargetSourceBuilder()
+        let (_, resources, _, _, _) = try builder.run()
```
freddi-kit, line 49:
```suggestion
        let resources = try builder.run().resources
```

I feel it is more readable

freddi-kit: ref https://github.com/apple/swift-package-manager/blob/main/Sources/PackageLoading/TargetSourcesBuilder.swift#L162

[warning] i did not check whether it can be done or not

Sources/ScipioKit/ProjectGenerator/ProjectGenerator.swift
```diff
@@ -220,6 +256,9 @@ class ProjectGenerator {
             ])
         )
 
+        // Make Resource target if needed
```
freddi-kit, line 259:
```suggestion
        // Make Resource target if needed (returns nil if not necessary)
```

we may expect this throw error if not needed, please add comment

Sources/ScipioKit/ProjectGenerator/BuildSettingsGenerator.swift
```diff
@@ -196,7 +197,47 @@ struct TargetBuildSettingsGenerator {
...
+        for dependencyModule in [target] + targetDependencies {
+            switch dependencyModule.underlyingTarget {
+            case let systemTarget as SystemLibraryTarget:
+                headerSearchPaths.append("$(SRCROOT)/\(systemTarget.path.relative(to: sourceRootDir).pathString)")
+            case let clangTarget as ClangTarget:
+                headerSearchPaths.append("$(SRCROOT)/\(clangTarget.includeDir.relative(to: sourceRootDir).pathString)")
+            default:
+                continue
+            }
+        }
```
mtj0928, line 214: [IMO] In this case, functional programming style might be better, because reader don't need tp take care that a value might be returned in the switch statement.
```swift

headerSearchPaths += ([target] + targetDependencies)
    .compactMap { dependencyModule in
        switch dependencyModule.underlyingTarget {
        case let systemTarget as SystemLibraryTarget:
            return "$(SRCROOT)/\(systemTarget.path.relative(to: sourceRootDir).pathString)"
        case let clangTarget as ClangTarget:
            return "$(SRCROOT)/\(clangTarget.includeDir.relative(to: sourceRootDir).pathString)"
        default:
            return nil
        }
    }
```

Sources/ScipioKit/ProjectGenerator/ProjectGenerator.swift
```diff
@@ -249,58 +288,126 @@ class ProjectGenerator {
...
-            let files = try walk(resource.path)
-            return try files.map { file in
-                try resourcesGroup.addFile(at: file.toPath(), sourceRoot: sourceRoot.toPath())
-            } + lists
+        let resourcesReferences: [PBXFileReference] = try resources.reduce([]) { files, resource in
+            let filePath = target.resourceDir.appendingPathComponent(resource.destination.pathString)
+            let file = try resourceTargetGroup.addFile(at: Path(filePath.path), sourceRoot: Path(target.resourceDir.path))
+            return files + [file]
         }
+        resourcesReferences.forEach { addObject($0) }
```
mtj0928, line 364: `map` is simpler than `reduce`, and we can move `addObject` into the closure.
```swift

let resourcesReferences: [PBXFileReference] = try resources.map { resource in
    let filePath = target.resourceDir.appendingPathComponent(resource.destination.pathString)
    let file = try resourceTargetGroup.addFile(at: Path(filePath.path), sourceRoot: Path(target.resourceDir.path))
    return addObject(file)
}
```

Sources/ScipioKit/ProjectGenerator/BuildSettingsGenerator.swift
```diff
@@ -196,7 +197,47 @@ struct TargetBuildSettingsGenerator {
...
+    }
+}
+
+struct ResourceBundleSettingsBuilder {
+    func generate(
+        for target: ResolvedTarget,
+        configuration: BuildConfiguration,
+        infoPlistPath: URL,
+        isSimulatorSupported: Bool
+    ) -> XCBuildConfiguration {
```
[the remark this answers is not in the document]
giginet: Other settings builder has members. So they are regular structs.
It's better to unify their usage.

Sources/ScipioKit/ProjectGenerator/BundleAccessorGenerator.swift
```diff
@@ -0,0 +1,43 @@
...
+extension Foundation.Bundle {
+    static let module: Bundle = {
+        let frameworkBundle: Bundle = .init(for: BundleMarker.self)
+        guard let resourceBundleURL = frameworkBundle.url(forResource: "\(resourceBundleName.asSwiftStringLiteralConstant)", withExtension: "bundle"),
+          let bundle = Bundle(url: resourceBundleURL) else {
+            fatalError("could not load resource bundle")
+        }
+        return bundle
+    }()
+}
```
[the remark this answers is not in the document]
giginet: We can separate this snippet into a template file.
However, it's difficult to fill substitutes.

I think it's better to embed it as a multiline string literal and use variables with String interpolation.

If this snippet is broken, we can notice by the E2E Tests (Failing to build resource modules)

Sources/ScipioKit/ProjectGenerator/ResourceCollector.swift
```diff
@@ -0,0 +1,53 @@
...
+            defaultLocalization: package.manifest.defaultLocalization,
+            additionalFileRules: FileRuleDescription.xcbuildFileTypes,
+            toolsVersion: package.manifest.toolsVersion,
+            fileSystem: TSCBasic.localFileSystem,
+            observabilityScope: observabilitySystem.topScope)
+    }
+
+    private let package: ResolvedPackage
+    private let target: ResolvedTarget
+    private let fileSystem: FileSystem
```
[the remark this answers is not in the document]
giginet: This variable provides for constructor injection. So it's better to put the default variable on its constructor.

Sources/ScipioKit/ProjectGenerator/ResourceCollector.swift
giginet: It works

giginet:
@mtj0928 @freddi-kit I fixed it. Could you review it again? https://github.com/giginet/Scipio/pull/22/commits/a747ec90f23ee875fad42c64650b3a53e80d2dc9

Sources/ScipioKit/ProjectGenerator/BuildSettingsGenerator.swift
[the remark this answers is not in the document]
mtj0928: In that case, how about removing members from other builders?
I think there is no strong reason why the struct has members.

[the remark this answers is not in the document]
giginet: Like ` ProjectBuildSettingsGenerator` needs the most context to generate build settings.
So it's a pain to pass every context by function arguments.
It's better to have members in the builder.

And these existing implementations are not related to this PR. So I'd like to fix this on the next PR for refactoring.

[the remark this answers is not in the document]
mtj0928: However, your idea is also OK for me.
So it's ok to ignore this comment.

Sources/ScipioKit/ProjectGenerator/ResourceCollector.swift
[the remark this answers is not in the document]
giginet: You mean about definition order.
I fixed.

Sources/ScipioKit/ProjectGenerator/BuildSettingsGenerator.swift
```diff
@@ -152,11 +153,11 @@ struct TargetBuildSettingsGenerator {
                 "PRODUCT_NAME": "$(TARGET_NAME:c99extidentifier)",
                 "PRODUCT_MODULE_NAME": "$(TARGET_NAME:c99extidentifier)",
                 "PRODUCT_BUNDLE_IDENTIFIER": .string(target.c99name.spm_mangledToBundleIdentifier()),
-                "SKIP_INSTALL": true,
```
giuk-jung, line 155: Why was this option removed?

giuk-jung (approved):
I can't review the details of Swift, but the general logic is LGTM

Sources/ScipioKit/ProjectGenerator/BuildSettingsGenerator.swift
```diff
@@ -81,6 +81,7 @@ struct ProjectBuildSettingsGenerator {
             "GCC_PREPROCESSOR_DEFINITIONS": [.inherited, "SWIFT_PACKAGE=1"],
             "USE_HEADERMAP": false,
             "CLANG_ENABLE_OBJC_ARC": true,
+            "SKIP_INSTALL": true,
```
freddi-kit, line 84: If my understand is correct, SKIP_INSTALL is false to build as xcframework, why it is set as true explicitly?

mtj0928 (approved):

Sources/ScipioKit/ProjectGenerator/BuildSettingsGenerator.swift
giginet: Your point is good. I'm confused about `SKIP_INSTALL` usage. As I investigated, `SKIP_INSTALL` is overwritten on `xcodebuild` command.
https://github.com/giginet/Scipio/blob/main/Sources/ScipioKit/XcodeBuild/Commands/ArchiveCommand.swift#L26

They should be more explicit.
I'll fix this.

giginet:
In further investigation, `SKIP_INSTALL` can be `NO` for all targets.
https://github.com/giginet/Scipio/pull/22/commits/399d66060dbd7b08f41a3207b202de5ce6a7b82a

So I specified `SKIP_INSTALL=NO` always clearly.

freddi-kit (approved):

PR #38, opened 2023-01-25
Introduce PIF based compiler

giginet:
https://github.com/giginet/Scipio/issues/36 closes #36

In the current implementation, we have to hassle to maintain the project generator.

I tried a new approach using [PIF](https://github.com/apple/swift-package-manager/blob/16c7cb679c88cc0bcc26483545aba12bffd5b8f9/Sources/XCBuildSupport/PIF.swift); it's an intermediate representation to talk SwiftPM and build systems.

PR #39, opened 2023-01-25
Remove FileSystem

giginet:
Remove Scipio.FileSystem and use `TSCBasic.FileSystem` instead.

it should be reviewed after merging #38


Sources/ScipioKit/Producer/BinaryExtractor.swift
```diff
@@ -1,6 +1,7 @@
 import Foundation
 import PackageGraph
 import PackageModel
+import TSCBasic
```
freddi-kit, line 4: Do we need it?

freddi-kit (approved):

PR #41, opened 2023-02-01
Update argument parser

giginet:
Currently, the dependency setting is broken.
I update swift-argument-parser to fix this


Package.resolved
```diff
@@ -41,17 +41,17 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-argument-parser.git",
       "state" : {
-        "revision" : "e394bf350e38cb100b6bc4172834770ede1b7232",
-        "version" : "1.0.3"
+        "revision" : "9f39744e025c7d377987f30b03770805dcb0bcd1",
+        "version" : "1.1.4"
```
giginet, line 45: Just execute `swift package update`

Sources/scipio/AsyncParsableCommand.swift
```diff
@@ -1,80 +0,0 @@
...
-// Copyright (c) 2020 Apple Inc. and the Swift project authors
-// Licensed under Apache License v2.0 with Runtime Library Exception
-//
-// See https://swift.org/LICENSE.txt for license information
-//
-// ===----------------------------------------------------------------------===//
-
-import ArgumentParser
-
-// This is just back-ported from a latest swift-argument-parser
```
giginet, line 14: This is backporting. So it should be removed.

freddi-kit (approved):

PR #44, opened 2023-02-07
Use upstream SwiftPM

giginet:
Currently, we use a fork. However, a required PR seems not to be accepted.

Use a public interface instead to avoid using forks.


giginet:
I found a bug on SwiftPM.

This PR helps it to fix this CI.
https://github.com/apple/swift-package-manager/pull/6130

giuk-jung (approved):

PR #52, opened 2023-02-19
Fix cacheMode always become disabled on CLI

giginet:
Due to latest changes, `cacheMode` will be always disabled


giginet:
Unfortunately, we don't have tests for CLI 🤕

freddi-kit (approved):

PR #54, opened 2023-02-27
Support request for cache storage concurrently

giginet:
In the current implementation, downloading and uploading cache runs sequentially.

It may slow for remote caching. They should be run concurrently.

This PR supports concurrent storage manipulation.
First, download all available cache, then build all targets without cache.


Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -74,94 +74,140 @@ struct FrameworkProducer {
...
-                try? await cacheSystem.cacheFramework(product, at: outputPath)
+        if isProducingCacheEnabled {
+            for target in targetsToBuild {
+                await generateVersionFile(for: target, using: cacheSystem)
+            }
+            await cacheSystem.cacheFrameworks(targetsToBuild)
+        }
+    }
 
-                if case .prepareDependencies = descriptionPackage.mode {
```
giginet, line 102: This condition was wrong.

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -74,94 +85,150 @@ struct FrameworkProducer {
...
-                )
-                try binaryExtractor.extract(of: binaryTarget, overwrite: overwrite)
-                logger.info("✅ Copy \(binaryTarget.c99name).xcframework", metadata: .color(.green))
-            default:
-                fatalError("Unexpected target type \(product.target.type)")
+    @discardableResult
+    private func buildXCFrameworks(
+        _ target: CacheSystem.CacheTarget,
+        outputDir: URL,
+        cacheSystem: CacheSystem
```
freddi-kit, line 200: Where is this cacheSystem used? I cannot find where `buildXCFrameworks` using it

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -74,94 +85,150 @@ struct FrameworkProducer {
...
+                            let restored = try await restore(target: target, cacheSystem: cacheSystem)
+                            return restored ? target : nil
+                        } catch {
+                            return nil
+                        }
+                    }
+                }
+                for await target in group.compactMap({ $0 }) {
+                    restored.insert(target)
+                }
```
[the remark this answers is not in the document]
giginet: In L151 is executed in each task.
So `restored` can't be shared because it causes race-condition.
All task attempt to write to `restored`

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -74,94 +87,152 @@ struct FrameworkProducer {
...
+                    logger.info("🗑️ Delete \(frameworkName)", metadata: .color(.red))
+                    try fileSystem.removeFileTree(outputPath.absolutePath)
+                }
+                let restoreResult = await cacheSystem.restoreCacheIfPossible(target: target)
+                switch restoreResult {
+                case .succeeded:
+                    logger.info("✅ Restore \(frameworkName) from cache storage", metadata: .color(.green))
+                    return true
+                case .failed:
+                    logger.warning("⚠️ Restoring \(frameworkName) is failed", metadata: .color(.yellow))
```
freddi-kit, line 191: Why not show detail error here?

Sources/ScipioKit/Producer/FrameworkProducer.swift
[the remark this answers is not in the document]
giginet: Added

Sources/ScipioKit/Producer/FrameworkProducer.swift
giginet: Added error feedback
https://github.com/giginet/Scipio/pull/54/commits/b2d6bc44d8a28bb32d1a567a0a90d06148b3388c

freddi-kit (approved):

PR #57, opened 2023-03-13
Fix order issue of `BuildOptions` #55

### LINKED ISSUE #55: Cache is not reused every time.
Scipio generates a version file for each dependency, and a cache is reused when the `CacheKey` of the generated file is equal.
However, `sdks` in a version file is a set and the order of `sdks` are not same every time.
As a result, the key will be changed.

The bellow version files represents the same version, but the order of `sdks` is not equal.
```
{
  "buildOptions" : {
    "enableLibraryEvolution" : true,
    "buildConfiguration" : "release",
    "isDebugSymbolsEmbedded" : true,
    "frameworkType" : "static",
    "sdks" : [
      "iOS",
      "iOSSimulator"
    ]
  },
  "targetName" : "Alamofire",
  "clangVersion" : "clang-1403.0.22.11",
  "pin" : {
    "version" : "5.6.4",
    "revision" : "78424be314842833c04bc3bef5b72e85fff99204"
  }
}
```
```
{
  "buildOptions" : {
    "enableLibraryEvolution" : true,
    "buildConfiguration" : "release",
    "isDebugSymbolsEmbedded" : true,
    "frameworkType" : "static",
    "sdks" : [
      "iOSSimulator",
      "iOS"
    ]
  },
  "targetName" : "Alamofire",
  "clangVersion" : "clang-1403.0.22.11",
  "pin" : {
    "version" : "5.6.4",
    "revision" : "78424be314842833c04bc3bef5b72e85fff99204"
  }
}
```

mtj0928:
The `sdk` of `BuildOptions` is `OrderedSet`, but some codes handles `sdk` as `Set<SDK>`.
As a result, the cache is sometimes not reused.

To fix the issue, I did
- Replacing most of `OrderedSet` with `Set`.
- Sorting the `sdk` when `BuildOptions` is created.

close #55

### DIFF
--- a/Sources/ScipioKit/BuildOptions.swift
+++ b/Sources/ScipioKit/BuildOptions.swift
@@ -6,7 +6,7 @@ struct BuildOptions: Hashable, Codable {
         buildConfiguration: BuildConfiguration,
         isDebugSymbolsEmbedded: Bool,
         frameworkType: FrameworkType,
-        sdks: OrderedSet<SDK>,
+        sdks: Set<SDK>,
         extraFlags: ExtraFlags?,
         extraBuildParameters: ExtraBuildParameters?,
         enableLibraryEvolution: Bool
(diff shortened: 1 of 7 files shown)


giginet (approved):
Great work 🌸

PR #58, opened 2023-03-15
Remove Resolver

giginet:
SwiftPM attempts to resolve when loading package manifests.

So `Resolver` is unnecessary.
It makes to improve setup speeds.


freddi-kit (approved):

Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -69,7 +65,8 @@ struct DescriptionPackage {
         let workspace = try Workspace(
             fileSystem: fileSystem,
             location: Workspace.Location(forRootPackage: packagePath, fileSystem: fileSystem),
-            configuration: workspaceConfiguration
+            configuration: workspaceConfiguration,
+            customHostToolchain: toolchain
```
giuk-jung, line 69: Is there some reason why `customHostToolchain` was added this time even though it is a optional customization?

giginet: Because, SwiftPM internal workspace tried to generate another user toolchain.
It should be shared.

PR #59, opened 2023-03-15
Remove addLink process to avoid duplicated symbols

giginet:
Screenshot 2023-03-15 at 17 01 14

These link settings are already added on SwiftPM's `PIFBuilder`.

So this process causes duplicated symbols when libraries are statics.

I checked in some situations this change will work fine.

### DIFF
--- a/Sources/ScipioKit/Producer/PIF/XCBuildClient.swift
+++ b/Sources/ScipioKit/Producer/PIF/XCBuildClient.swift
@@ -108,7 +108,6 @@ struct XCBuildClient {
                 to: destination
             )
         }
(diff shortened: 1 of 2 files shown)


Sources/ScipioKit/Producer/PIF/XCBuildClient.swift
```diff
@@ -108,7 +108,6 @@ struct XCBuildClient {
                 to: destination
             )
         }
-        print(modulesDir.appending(component: "module.modulemap"))
```
giginet, line 111: 😅

freddi-kit (approved):

giuk-jung (approved):

PR #66, opened 2023-05-11
Combine scipio-s3-storage

giginet:
Combine [scipio-s3-storage](https://github.com/giginet/scipio-s3-storage) into this repo to become a mono-repo. 

Scipio and scipio-s3-storage depends on SwiftPM, unfortunately, SwiftPM doesn't provide semantic-versioning.
So it can't use a version specifier.
As a result, it's very inconvenient to manage dependency on using scipio-s3-storage.


Sources/ScipioS3Storage/Compressor.swift
```diff
@@ -0,0 +1,81 @@
+import Foundation
+import System
+import Compression
+import AppleArchive
+
+struct Compressor {
```
freddi-kit, line 6: The name of `Compressor` is a little ambigious so please rename like `AARArchiver` and add comment about AAR

Sources/ScipioS3Storage/Compressor.swift
```diff
@@ -0,0 +1,81 @@
...
+    private let fileManager: FileManager = .default
+
+    init() {
+        try? fileManager.createDirectory(at: temporaryDirectory, withIntermediateDirectories: true)
+    }
```
freddi-kit, line 11: throw error is better

Sources/ScipioS3Storage/Compressor.swift
```diff
@@ -0,0 +1,81 @@
...
+    func extract(_ archiveData: Data, to destinationPath: URL) throws {
+        let destination = FilePath(destinationPath.path)
+
+        let temporaryPath = temporaryDirectory.appendingPathComponent("\(UUID().uuidString).aar")
+        fileManager.createFile(atPath: temporaryPath.path, contents: archiveData)
+        defer { try? fileManager.removeItem(at: temporaryPath) }
+
+        try fileManager.createDirectory(at: destinationPath, withIntermediateDirectories: true)
+
+        _ = try ArchiveByteStream.withFileStream(
```
freddi-kit, line 50: Isn't its return value discardable value?

giginet: This returns a return value of closure.
https://developer.apple.com/documentation/applearchive/archivebytestream/3588517-withfilestream

It is not needed here because closure writes this to the file.

freddi-kit (approved):

PR #67, opened 2023-05-11
Add scipioVersion to VersionFile

giginet:
Add scipioVersion key to VersionFile.
It helps to re-generate XCFrameworks when the compiler is updated.


Plugins/GenerateScipioVersion/plugin.swift
```diff
@@ -0,0 +1,56 @@
...
+
+@main
+struct GenerateScipioVersion: BuildToolPlugin {
+    func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
+        let zshPath = Path("/bin/zsh") // execute dummy command
+        let generatedSourceDir = context.pluginWorkDirectory
+        let generatedSourcePath = generatedSourceDir
+            .appending(subpath: "ScipioVersion.generated.swift")
+
+        let revision = try? fetchRepositoryVersion(context: context)
```
freddi-kit, line 12: Why can we ignore this error by `try?`?

Plugins/GenerateScipioVersion/plugin.swift
```diff
@@ -0,0 +1,56 @@
...
+            "rev-parse",
+            "HEAD",
+        ]
+        process.standardOutput = standardOutput
+
+        try process.run()
+        process.waitUntilExit()
+        let outputData = try standardOutput.fileHandleForReading.readToEnd()
+        guard let outputData else { return nil }
+        let revision = String(data: outputData, encoding: .utf8)?
```
freddi-kit, line 52: How about `fatalError` if the revision is nil?

Plugins/GenerateScipioVersion/plugin.swift
giginet: This external command execution is in the build phase by a plugin.
So it fails silently. This is why to use `try?`

if failed, scipioVersion is not filled on VersionFile, `scipioVersion` is nullable.

Plugins/GenerateScipioVersion/plugin.swift
giginet: Same reason above

freddi-kit (approved):
LGTM, r? @giuk-jung

PR #69, opened 2023-05-25
Polishing English expression

iXerol:
"In default" in the README should be correctly expressed as "by default".


giginet (approved):
Thanks 😄

PR #94, opened 2023-08-01
Fix moduleMap generation for umbrellaDirectory

giginet:
If the moduleMap types are detected as `umbrellaDirectory`, `publicHeaderPath` would be embedded into modulemaps distributed in frameworks.

### Before

Generated module map from the package with `umbrellaDirectory` type has an `umbrella` declare

```
framework module MyTarget {
    umbrella "include"
    export *
}
```

However, when headers are copied to frameworks, directory structures are lost.
So it would not work well.

### After

This PR makes to change a generating modulemaps.

It expands directory contents recursively.

```
framework module MyTarget {
    header "a.h"
    header "b.h"
    export *
}
```


giuk-jung:
LGTM. It seems to be fine, at least in the configurations I've experienced.

giuk-jung (approved):

freddi-kit (approved):

PR #105, opened 2023-09-27
Support both Swift 5.8 and Swift 5.9

giginet:
This PR migrate the default Xcode version to 5.9.
I also added `Package@swift-5.8.swift` to support backward-compatibility.
Run tests both on Xcode 14.3 and Xcode 15.0


Package@swift-5.8.swift
```diff
@@ -0,0 +1,94 @@
+// swift-tools-version: 5.6
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+import Foundation
+
+let package = Package(
+    name: "Scipio",
+    platforms: [
+        .macOS(.v12),
```
freddi-kit, line 10: How about change it to v13 in  both Package.swift side? Many engineer may use new macOS

Sources/ScipioKit/Producer/PIF/ToolchainGenerator.swift
```diff
@@ -39,12 +42,24 @@ struct ToolchainGenerator {
...
+        #else
+
         return Destination(
             hostTriple: try? Triple("arm64-apple-\(sdk.settingValue)"),
             targetTriple: try? Triple("arm64-apple-\(sdk.settingValue)"),
             sdkRootDir: sdkPath,
             toolchainBinDir: toolchainDirPath,
             extraFlags: BuildFlags(cCompilerFlags: extraCCFlags, swiftCompilerFlags: extraSwiftCFlags)
         )
+        #endif
```
freddi-kit, line 63: What is changed here? Do we need to use variable in swift 5.9 here?

Package.swift
```diff
@@ -20,8 +20,8 @@ let package = Package(
             targets: ["ScipioS3Storage"]),
     ],
     dependencies: [
-        .package(url: "https://github.com/giginet/swift-package-manager.git",
-                 revision: "11c88e87e6874a570711d18462ca0ea43ff7ff5e"),
+        .package(url: "https://github.com/apple/swift-package-manager.git",
```
freddi-kit, line 23: 🎉  your changes released, right?

freddi-kit (approved):

Sources/ScipioKit/Producer/PIF/ToolchainGenerator.swift
giginet: 🤔

giginet: I re-checked that this was unnecessary. So I reverted.

Package@swift-5.8.swift
giginet: I'll upgrade the development environment later because I want to remove NSRegularExpression. But I'll do this next time.

Package@swift-5.8.swift
```diff
@@ -0,0 +1,94 @@
...
+        .library(
+            name: "ScipioKit",
+            targets: ["ScipioKit"]),
+        .library(
+            name: "ScipioS3Storage",
+            targets: ["ScipioS3Storage"]),
+    ],
+    dependencies: [
+        .package(url: "https://github.com/giginet/swift-package-manager.git",
+                 revision: "99575e7a3eb8e66112ec45e2f11ef23a1828b110"),
```
giginet, line 24: I upgrade the fork to the commit https://github.com/giginet/swift-package-manager/commit/99575e7a3eb8e66112ec45e2f11ef23a1828b110

SwiftPM refers HEAD of swift-driver and swift-support-tools-core.
Current HEAD of them only supports above macOS 12.0. But old SwiftPM fork still supports legacy macOS.
This causes a resolving error. So I bump up supported version on our fork.

This fork should be used for legacy Xcode users (14.3). So it will be removed after dropping 14 support.

Package.resolved
```diff
@@ -1,238 +0,0 @@
-{
```
giginet, line 1: I removed `Package.resolved`. Because it will be changed by resolving with various Xcode versions.

freddi-kit (approved):

PR #106, opened 2023-09-29
Add option to enable manifest update closes #97

### LINKED ISSUE #97: Add option to enable update manifest
Currently, Scipio doesn't update `Package.resolved`. However, is some cases, this behavior could be more convenient.

It's better to add new option to configure SwiftPM resolver behavior.

The option is the same with `--force-resolved-versions, --disable-automatic-resolution` of SwiftPM.

giginet:
closes #97

Introduce new option `--only-use-versions-from-resolved-file`.

Enabling this option, SwiftPM never updated `Package.resolved` automatically.
If `Package. resolved` and the existing workspace is mismatched, resolving will fail.

Since this PR, the default behavior is changed. The default value becomes `false`.
Because it's inconvenient to enable this for common use cases.

### DIFF
--- a/Sources/ScipioKit/DescriptionPackage.swift
+++ b/Sources/ScipioKit/DescriptionPackage.swift
@@ -78,7 +78,13 @@ struct DescriptionPackage {
         return workspace
     }
 
-    init(packageDirectory: AbsolutePath, mode: Runner.Mode) throws {
+    /// Make DescriptionPackage from a passed package directory
+    /// - Parameter packageDirectory: A path for the Swift package to build
+    /// - Parameter mode: A Scipio running mode
+    /// - Parameter onlyUseVersionsFromResolvedFile: A boolean value indicates disabling force updating of Package.resolved.
+    ///   If it is `true` Package.resolved never be updated
+    ///   instead the resolving will fail if the Package.resolved is mis-matched with the workspace.
+    init(packageDirectory: AbsolutePath, mode: Runner.Mode, onlyUseVersionsFromResolvedFile: Bool) throws {
         self.packageDirectory = packageDirectory
         self.mode = mode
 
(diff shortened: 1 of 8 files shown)


Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -78,7 +78,13 @@ struct DescriptionPackage {
         return workspace
     }
 
-    init(packageDirectory: AbsolutePath, mode: Runner.Mode) throws {
+    /// Make DescriptionPackage from a passed package directory
+    /// - Parameter packageDirectory: A path for the Swift package to build
+    /// - Parameter mode: A Scipio running mode
+    /// - Parameter onlyUseVersionsFromResolvedFile: A boolean value indicates disabling force updating of Package.resolved.
+    ///   If it is `true` Package.resolved never be updated
+    ///   instead the resolving will fail if the Package.resolved is mis-matched with the workspace.
```
freddi-kit, line 86: It hards to read as sentence since there is no , for if.

```suggestion
    ///   If it is `true`, Package.resolved never be updated.
    ///   Instead, the resolving will fail if the Package.resolved is mis-matched with the workspace.
```

Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -78,7 +78,13 @@ struct DescriptionPackage {
         return workspace
     }
 
-    init(packageDirectory: AbsolutePath, mode: Runner.Mode) throws {
+    /// Make DescriptionPackage from a passed package directory
+    /// - Parameter packageDirectory: A path for the Swift package to build
+    /// - Parameter mode: A Scipio running mode
+    /// - Parameter onlyUseVersionsFromResolvedFile: A boolean value indicates disabling force updating of Package.resolved.
```
freddi-kit, line 84: Could you explain what `only` means?

giginet, line 84:
```suggestion
    /// - Parameter onlyUseVersionsFromResolvedFile: A boolean value indicates disabling force updating of Package.resolved.
    /// If true, use package versions only from existing Package.resolved.
```

giginet, line 84:
```suggestion
     /// - Parameter onlyUseVersionsFromResolvedFile: A boolean value if true disabling force updating of Package.resolved.
    /// Then, use package versions only from existing Package.resolved.
```

freddi-kit (approved):

PR #119, opened 2024-05-10
Pass buildOptionsMatrix to each target

giginet:
Fix the issue that `buildOptionsMatrix` sometimes doesn't work for the specific target.

## Description

Scipio generates `PIF` for each package. Each target in the package uses the same PIF. However, `PIFGenerator` doesn't refer to the matrix, so target settings were applied for all targets in the same package.

This PR fixes this issue to refer to the matrix in generating the PIF phase.


giuk-jung (approved):
> This PR fixes this issue to refer to the matrix in generating the PIF phase.

LGTM, I've confirmed that the code has been modified to generate the correct PIF.

PR #120, opened 2024-05-15
Support strict-concurrency

giginet:
Enable `strict-concurrency` flag for each target.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,8 +1,13 @@
-// swift-tools-version: 5.6
+// swift-tools-version: 5.10
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
 import Foundation
+
+let swiftSettings: [SwiftSetting] = [
+    .enableExperimentalFeature("StrictConcurrency"),
+    .unsafeFlags(["-strict-concurrency=complete"]),
+]
 
 let package = Package(
     name: "Scipio",
@@ -28,21 +33,24 @@ let package = Package(
                  from: "1.0.4"),
         .package(url: "https://github.com/apple/swift-argument-parser.git",
                  from: "1.1.0"),
-        .package(url: "https://github.com/apple/swift-algorithms.git", 
+        .package(url: "https://github.com/apple/swift-algorithms.git",
                  from: "1.0.0"),
         .package(url: "https://github.com/onevcat/Rainbow",
                  .upToNextMinor(from: "4.0.1")),
-        .package(url: "https://github.com/soto-project/soto-codegenerator", 
+        .package(url: "https://github.com/soto-project/soto-codegenerator",
                  from: "0.6.0"),
-        .package(url: "https://github.com/soto-project/soto-core.git", 
+        .package(url: "https://github.com/soto-project/soto-core.git",
                  from: "6.4.0"),
     ],
     targets: [
-        .executableTarget(name: "scipio",
-                          dependencies: [
-                            .target(name: "ScipioKit"),
-                            .product(name: "ArgumentParser", package: "swift-argument-parser"),
-                          ]),
+        .executableTarget(
+            name: "scipio",
+            dependencies: [
+                .target(name: "ScipioKit"),
+                .product(name: "ArgumentParser", package: "swift-argument-parser"),
+            ],
+            swiftSettings: swiftSettings
+        ),
         .target(
             name: "ScipioKit",
             dependencies: [
@@ -53,12 +61,13 @@ let package = Package(
                 .product(name: "Algorithms", package: "swift-algorithms"),
                 .product(name: "Rainbow", package: "Rainbow"),
             ],
+            swiftSettings: swiftSettings,
             plugins: [
                 .plugin(name: "GenerateScipioVersion")
             ]
         ),
         .plugin(
-            name: "GenerateScipioVersion", 
+            name: "GenerateScipioVersion",
             capability: .buildTool()
         ),
         .target(
@@ -88,7 +97,7 @@ let isDevelopment = ProcessInfo.processInfo.environment["SCIPIO_DEVELOPMENT"] ==
 
 // swift-docs is not needed for package users
 if isDevelopment {
-    package.dependencies += [ 
+    package.dependencies += [
         .package(url: "https://github.com/apple/swift-docc-plugin", from: "1.3.0"),
     ]
 }
--- a/Package.swift
+++ b/Package.swift
@@ -12,7 +12,7 @@ let swiftSettings: [SwiftSetting] = [
 let package = Package(
     name: "Scipio",
     platforms: [
-        .macOS(.v12),
+        .macOS(.v13),
     ],
     products: [
         .executable(name: "scipio",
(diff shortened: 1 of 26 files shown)


Package.swift
```diff
@@ -1,9 +1,14 @@
-// swift-tools-version: 5.6
+// swift-tools-version: 5.10
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
 import Foundation
 
+let swiftSettings: [SwiftSetting] = [
+    .enableExperimentalFeature("StrictConcurrency"),
+    .unsafeFlags(["-strict-concurrency=complete"]),
```
mtj0928, line 9: I'm not sure, but is the unsafeFlags necessary?

Sources/ScipioKit/Logger.swift
```diff
@@ -68,26 +68,26 @@ struct ScipioLogHandler: LogHandler {
...
-    switch diagnostics.severity {
-    case .error:
-        logger.error("\(diagnostics.message)")
-    case .warning:
-        logger.warning("\(diagnostics.message)")
-    case .info:
-        logger.info("\(diagnostics.message)")
-    case .debug:
-        logger.debug("\(diagnostics.message)")
+func generateObservabilitySystem() -> ObservabilitySystem {
```
mtj0928, line 80: `generate` is a bit unnatural on Swift language for making an instance.
And I doubt this function should be located at global level.

How about this?
```swift
extension ObservabilitySystem {
    static func makeForScipio() -> ObservabilitySystem {
        Observability { .... }
    }
}
```

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -148,28 +148,37 @@ struct CacheSystem {
...
     func cacheFrameworks(_ targets: Set<CacheTarget>) async {
         let chunked = targets.chunks(ofCount: storage?.paralellNumber ?? CacheSystem.defaultParalellNumber)
 
         for chunk in chunked {
             await withTaskGroup(of: Void.self) { group in
                 for target in chunk {
+                    let context = TaskContext(
+                        outputDirectory: outputDirectory,
+                        frameworkName: target.buildProduct.frameworkName
+                    )
```
mtj0928, line 176: Is this context required?
`URL` and `CacheTarget` are Sendable, so I think we can directory use them in `addTask`.

Tests/ScipioKitTests/RunnerTests.swift
```diff
@@ -297,7 +300,7 @@ final class RunnerTests: XCTestCase {
         XCTAssertTrue(fileManager.fileExists(atPath: storageDir.appendingPathComponent("ScipioTesting").path))
 
         addTeardownBlock {
-            try self.fileManager.removeItem(at: storageDir)
+            try FileManager.default.removeItem(at: storageDir)
```
mtj0928, line 303:
```swift
addTeardownBlock { [fileManager] in 
    try fileManager.removeItem(at: storageDir)
}
```

Package.swift
giginet: Yes, I also thought it's fine only with `ExperimentalFeature` but it seems to need the `unsafeFlags` too.

Sources/ScipioKit/Logger.swift
giginet: It's a bit in-convenient because callers must imports `Basics` module. It sometimes occurs conflicts of some symbols.

Then I renamed it to `makeObservabilitySystem`

Tests/ScipioKitTests/RunnerTests.swift
giginet: I tried this, but `FileManager` is a non-sendable class, so we can't capture them for the `@Sendable` closures.

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -159,26 +159,19 @@ struct CacheSystem: Sendable {
...
         let chunked = targets.chunks(ofCount: storage?.paralellNumber ?? CacheSystem.defaultParalellNumber)
 
         for chunk in chunked {
             await withTaskGroup(of: Void.self) { group in
                 for target in chunk {
-                    let context = TaskContext(
-                        outputDirectory: outputDirectory,
-                        frameworkName: target.buildProduct.frameworkName
-                    )
+                    let outputDirectory = outputDirectory
```
mtj0928, line 168: This local variable definition is meaningless./

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -170,45 +167,55 @@ struct FrameworkProducer {
...
+        func restore(
+            target: CacheSystem.CacheTarget,
+            cacheSystem: CacheSystem
+        ) async throws -> Bool {
+            let product = target.buildProduct
+            let frameworkName = product.frameworkName
+            let outputPath = outputDir.appendingPathComponent(product.frameworkName)
+            let exists = fileSystem.exists(outputPath.absolutePath)
+
+            if cacheMode.isConsumingCacheEnabled {
```
mtj0928, line 186: This comment is out of scope of this PR, but an early return is better in this case.
```swift
if !cacheMode.isConsumingCacheEnabled {
    return false
}

let expectedCacheKey = try await cacheSystem.calculateCacheKey(of: target)
// ...
```

(Please address this comment in another PR.)

Tests/ScipioKitTests/RunnerTests.swift
mtj0928: Hmm...
The current code is inconsistent, because there are similar code, `FileManager.default` and `fileManager`, in the test code. 
And more, the test code depends on the assumption that `fileManger` is `FileManager.default`, and it should be avoided.

There are three options.
1. Replace all `fileManager` with `FileManager.default`.
2. Stop using `addTeardownBlock` and use just `defer` for the clean up.
3. Add `unchecked Sendable` to the closure. This is test codes, so I expect there are no major issues related to data races. (However, we should not use it as much as possible). 

What do you think?

Sources/ScipioKit/Producer/PIF/BuildParametersGenerator.swift
```diff
@@ -3,8 +3,8 @@ import TSCBasic
 import XCBuildSupport
 import SPMBuildCore
 
-struct XCBBuildParameters: Encodable {
-    struct RunDestination: Encodable {
+struct XCBBuildParameters: Encodable, Sendable {
```
giuk-jung, line 6: Just a question: what was the reason for adding `Sendable`?

giuk-jung (approved):
LGTM. I've reviewed what I can.

Sources/ScipioKit/Producer/PIF/BuildParametersGenerator.swift
giginet: I don't remember the details, but it's used somewhere.

This struct is not public and struct. So it may be `Sendable` implicitly, so we might not specify this protocol.

Tests/ScipioKitTests/RunnerTests.swift
giginet: In final, I removed `defer` / `addTeardownBlock` for these cases.
It's enough to call on the bottom of methods.

mtj0928 (approved):

mtj0928 (approved):

PR #121, opened 2024-05-20
Add Xcode Version to CacheKey

giginet:
This PR adds Xcode version to the cache key.

Currently, we are using `clang --version` to detect the compiler versions are changed.
However, some Xcode releases changed SDK versions even the Clang version was not changed.

In this case, wrong binaries are used in the cache system.

This change may avoid the situation.

Since this PR, macOS 13 or above is required.


freddi-kit (approved):

Package@swift-5.9.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.6
+// swift-tools-version: 5.8
```
giuk-jung, line 1:
```suggestion
// swift-tools-version: 5.9
```

giginet: We haven't dropped 5.8 support yet. So I set the lowest version

giuk-jung (approved):

Sources/ScipioKit/Producer/Cache/XcodeVersion.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import RegexBuilder
+
+public struct XcodeVersion: Sendable, Hashable, Codable {
+    var xcodeVersion: String
+    var xcodeBuildVersion: String
```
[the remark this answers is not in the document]
giginet: Perhaps it's enough to identify the key only with the build version.
However, it's not a human-friendly. `xcodeBuildVersion` is good for human readability so it's better to contain both.

PR #129, opened 2024-06-16
Fix log level

mtj0928:
`scipio` command doesn't show info log when `verbose`  is not specified from https://github.com/giginet/Scipio/commit/9fb11bf27ae650576edccafb6696012ba096af09 .
I asked @giginet about the issue in person, and he said it's a bug.

In this PR, I fixed the log level.

### DIFF
--- a/Sources/scipio/CreateCommands.swift
+++ b/Sources/scipio/CreateCommands.swift
@@ -21,7 +21,7 @@ extension Scipio {
         var platforms: [Runner.Options.Platform] = []
 
         mutating func run() async throws {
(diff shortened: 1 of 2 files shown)


mtj0928:
Original logic in 18.3.0.
https://github.com/giginet/Scipio/blob/f9a3acc02c8e260ad3e55770d23776ed5646c44c/Sources/ScipioKit/Runner.swift#L38-L42

giginet (approved):
Thanks! I'll release a patch version soon.

Issue #130, opened 2024-06-18
Proposal: Replace `OrderedSet` with `Set`


dmhts:
Hey @giginet,

First of all, I wanted to thank you for making such a great tool and for keeping working on it.

In the code snippets below you can see that `allTargets` is an `OrderedSet` which leads to issues when `Scipio` is repeatedly used over a big set of dependencies with the `.storage(LocalCacheStorage(), ...)` cache mode.

https://github.com/giginet/Scipio/blob/f332eb4d30db7f2bb8ed3c75ccc0be53b0a88b53/Sources/ScipioKit/Producer/FrameworkProducer.swift#L92-L94

https://github.com/giginet/Scipio/blob/f332eb4d30db7f2bb8ed3c75ccc0be53b0a88b53/Sources/ScipioKit/Producer/FrameworkProducer.swift#L119

Namely, it leads to a crash in the `OrderedSet` implementation due to a failing assertion:

https://github.com/apple/swift-collections/blob/c60d2f274d61550ca7a80b5306d23235c5f5f85b/Sources/OrderedCollections/OrderedSet/OrderedSet%2BInsertions.swift#L22-L25

I'm not sure if duplicates can be avoided in your implementation, but the simplest way to solve the issue would be to use an unordered `Set` instead of the ordered one. I can confirm that making that change solves the issue. 

Should I open a PR to fix that or do you see any reasons to keep the set ordered?

giginet:
Thank you for the reporting issue!

As I recall, order was more important than duplication. Some list represents the dependency tree, so order have to be kept to build in the correct order.

Of course, I'm welcome to open PR, or I'll reduce `OrderSet` usage as much as possible.

omochi:
これ僕が実装した部分ですね。

ここがOrderedSetになっているのは、依存関係をグラフソートした結果を維持しているからです。
こうしないとビルド順が不適切になり問題が起きていたためです。

仮に全てをPIFで適切に定義できれば、どんな順番でビルドしても、xcodebuild側で依存解決するので動きますが、
その制約を導入するのは実装と保守が難しくなるでしょう。

一方、キャッシュのキーとして順序が不要というのはそうかもしれません。
対応するのであれば、ここの OrderedSet はそのままにしつつ、
キャッシュキーの構築をするところだけ Set にして並び順の情報を落とす、
という対応が望ましいでしょう。

---

This is the part I implemented.

[...]

PR #132, opened 2024-07-03
Fix: Duplicate build products in swift-6.0-DEVELOPMENT-SNAPSHOTs

dmhts:
### Motivation:
For our current project, we need both `Scipio` and some new APIs from the SwiftPM 6.0 latest development snapshot. As `SwiftPM` is also a direct `Scipio` dependency, I adapted `Scipio` to the changes introduced in the development snapshot in [our fork](https://github.com/dmhts/Scipio/commit/9d52daee2c7d3e92641b5e9e5996e394ef79a27e). All of them are trivial so no problems with that.

At the same time, during the development process, I started facing crashes related to violating the `Hashable` protocol requirements, which I've already mentioned [here](https://github.com/giginet/Scipio/issues/130) and [here](https://github.com/giginet/Scipio/pull/131#issuecomment-2198475596). It turned out that all of these issues are related to a refactoring introduced in the latest SwiftPM development snapshot that will be included in the Swift 6.0 release. Specifically, the `Hashable` conformance has been made unavailable for both [ResolvedModule](https://github.com/swiftlang/swift-package-manager/blob/f61175add974588b7e18174b88a4e94e23f9d9da/Sources/PackageGraph/Resolution/ResolvedModule.swift#L332-L333) and [PackageResolved](https://github.com/swiftlang/swift-package-manager/blob/123fa36ee4b4e778e57060d6c9e66f199158f649/Sources/PackageGraph/Resolution/ResolvedPackage.swift#L100-L101).

The changes above make it important to properly conform `BuildProduct` to `Hashable`, as it incorporates both types mentioned above. This PR addresses that by preparing the codebase for future changes, alleviating some migration hurdles that will need to be addressed in the coming months, all without breaking any existing functionality.

The changes in this PR fix all the non-deterministic crashes related to `Set` operations where `BuildProduct` is involved when compiled against the latest SwiftPM development snapshot. Additionally, the refactorings in SwiftPM mentioned above cause Scipio to produce duplicate targets. This issue is addressed by the unit tests, and it's also related to the crashes mentioned above.

Are you already planning migration to the latest SwiftPM versions and what do you think overall about this PR?


Tests/ScipioKitTests/DescriptionPackageTests.swift
```diff
@@ -74,10 +74,12 @@ final class DescriptionPackageTests: XCTestCase {
...
 
         XCTAssertEqual(
-            Set(try package.resolveBuildProducts().map(\.target.name)),
+            try package.resolveBuildProducts().map(\.target.name),
             [
                 "Logging",
                 "TestingPackage",
+                "ExecutableTarget",
+                "MyPlugin"
             ]
```
dmhts, line 83: If you comment out the custom Hashable conformance above this test will fail when Scipio depends on the latest SwiftPM development snapshot.

dmhts:
I'll have a look into the linting issue and the failing tests tomorrow.

giginet:
Thank you! I'll check this.

By the way, we're maintaining a fork it uses 6 snapshot. It may help you.
https://github.com/giginet/Scipio/tree/swift6

We're going to merge this into main to maintain both for Swift 5.10 and Swift 6 compiler

Tests/ScipioKitTests/Resources/Fixtures/TestingPackage/Package.swift
```diff
@@ -10,20 +10,38 @@ let package = Package(
...
             ]
         ),
+        .plugin(
+            name: "MyPlugin",
+            capability: .command(
+                intent: .custom(
+                    verb: "my-plugin-verb",
+                    description: "my-plugin-description")),
+            dependencies: ["ExecutableTarget"]
+        ),
```
giginet, line 38: In my understood, SwiftPM doesn't support Plugin and Macros for Xcode build mode.
https://github.com/swiftlang/swift-package-manager/blob/release/6.0/Sources/XCBuildSupport/PIFBuilder.swift#L406-L411

I wonder it works well

Tests/ScipioKitTests/DescriptionPackageTests.swift
```diff
@@ -74,10 +74,12 @@ final class DescriptionPackageTests: XCTestCase {
         XCTAssertEqual(package.name, "TestingPackage")
 
         XCTAssertEqual(
-            Set(try package.resolveBuildProducts().map(\.target.name)),
+            try package.resolveBuildProducts().map(\.target.name),
```
giginet, line 77: By using `Array` instead of `Set`, this must be care about their orders. How about adding `sorted` instead?

giginet (commented):
Thank you for explaining the context.

I think it makes sense to use `target.name` and `package.identifier` to represent their identity.
It will be merged after resolving CI errors.

I'm developing Swift 6 support in other branch. I'll merge these implementation into main in the near future. Please moment.

giginet:
As I saw the Swift 6 snapshot implementations, `IdentifiableSet` is widely used for this usage, so it may be better to conform to `Identifiable` instead of `Hashable`.

Tests/ScipioKitTests/Resources/Fixtures/TestingPackage/Package.swift
dmhts: Certainly, plugins and macros are not meant to be built. However, we're talking here about the resolving phase where plugins may depend on executable targets, which in turn may depend on regular targets already intended to be built.

Correct me if I'm wrong:
There are two phases:
1. Resolve phase: all types of targets are important since any target may have a dependency on a regular target (which we are trying to identify while resolving).
2. Build phase: before starting, all irrelevant types of targets (everything except `.library` and `.binary`) [are filtered out anyway](https://github.com/giginet/Scipio/pull/132).

Is that right?

dmhts: You can see an example in the [swift-argument-parser](https://github.com/apple/swift-argument-parser/blob/main/Package.swift):

`GenerateManual (plugin)` -> `generate-manual(executableTarget)` -> `ArgumentParser(regularTarrget)` and `ArgumentParserToolInfo(regularTarget)`

dmhts:
@giginet 
> As I saw the Swift 6 snapshot implementations, IdentifiableSet is widely used for this usage, so it may be better to conform to Identifiable instead of Hashable.

[This is the PR](https://github.com/swiftlang/swift-package-manager/pull/7248) that introduced `IdentifiableSet` and as you can see from the description the only reason for doing so was to improve performance for [the deep package graph loading cases (with 100 packages)](https://github.com/swiftlang/swift-package-manager/blob/ea3461d18808c2afeaab19b375f7838a5e540e2e/Tests/PackageGraphPerformanceTests/PackageGraphPerfTests.swift#L27). I doubt it's going to be a problem for Scipio any time soon as it works on a per-package basis. I just measured how much time it takes to resolve build products for `swift-argument-parser` and it's about `2.2ms`. I think we can survive that :)

```swift
let startTime = CACurrentMediaTime()
let targets = try descriptionPackage.resolveBuildProducts()
print("Runtime: \((CACurrentMediaTime() - startTime) * 1000) ms")
// Runtime: 2.2692499915137887 ms
```

On the other hand, conforming `BuildProduct` to `IdentifiableSet` instead of `Hashable` would bring some inconveniences, such as:
- `topologicalSort<T: Hashable>` requires `Hashable`, and this is the reason `ResolvedModule.Dependency` still conforms to `Hashable` as [you can see here](https://github.com/swiftlang/swift-package-manager/blob/ea3461d18808c2afeaab19b375f7838a5e540e2e/Sources/PackageGraph/Resolution/ResolvedModule.swift#L305-L314).
- other hashable types such as `CacheTarget` would require calculating a hash from `BuildProduct` in one way or another
- there are probably other possible issues that I cannot see right now.

In the end it's your call if you still want to use `IdentifiableSet` or not.

dmhts:
@giginet Have you considered using any output beautifiers such as [these](https://docs.fastlane.tools/best-practices/xcodebuild-formatters/)? At the moment it's extremely hard to understand what part of the integration test fails.

Tests/ScipioKitTests/DescriptionPackageTests.swift
dmhts: Isn't the idea also to check the order? `resolveBuildProducts` sorts all targets topologically and then reverses the order so that the build process can start from the leaf nodes. If I understand correctly, shouldn't the order also be preserved in the test?

```swift
[
                "Logging", // The leaf node
                "MyTarget", // Depends only on "Logging"
                "ExecutableTarget", // Depends only on "MyTarget"
                "MyPlugin", // Depends only on "ExecutableTarget", also the top package exposed via "products"
            ]
```

giginet:
I don't prefer to use external dependencies as possible so I'd like to avoid using `xcbeautify` or `xcpretty`.

Additionally, we run the tests using `swift test`. These formatters may support only xcodebuild.

giginet:
Thank you for measuring the performance.

As you said, it's no problem to use `Hashable` now.
I'm going to keep compatibility with 5.10 Snapshot for a while, so `IdentifiableSet` can't be used yet.

However, it's better to follow the implementations with the official one, so we'll migrate all `Set` to `IdentifiableSet` after dropping 5.10 support.

Tests/ScipioKitTests/DescriptionPackageTests.swift
giginet: oops,  I misunderstood, the order must be kept. It's great to add comments like your comment.

Tests/ScipioKitTests/Resources/Fixtures/TestingPackage/Package.swift
giginet: I see. You meant you want to make sure that the products are resolved in the correct order, regardless of whether it is built or not.

Your change seems to make sense.

giginet (approved):
I understand your context. Thank you for explaining!

Tests/ScipioKitTests/DescriptionPackageTests.swift
```diff
@@ -74,10 +74,12 @@ final class DescriptionPackageTests: XCTestCase {
...
             [
                 "Logging",
-                "TestingPackage",
+                "MyTarget",
+                "ExecutableTarget",
+                "MyPlugin",
             ]
         )
```
giginet, line 84:
```suggestion
            ],
            "Order of the resolved products should be correct"
        )
```

dmhts:
> I don't prefer to use external dependencies as possible so I'd like to avoid using xcbeautify or xcpretty.
> Additionally, we run the tests using swift test. These formatters may support only xcodebuild.

Valid point, just a note that [xcbeautify supports Swift Package Manager output](https://github.com/cpisciotta/xcbeautify?tab=readme-ov-file#features).

dmhts:
> However, it's better to follow the implementations with the official one, so we'll migrate all Set to IdentifiableSet after dropping 5.10 support.

Also a totally valid point, just make sure that the `IdentifiableSet` conformance excludes the `BuildTriple` from the identity definition, as it would result in duplicates in the output. In any case, there is already a test to back you up.

Tests/ScipioKitTests/DescriptionPackageTests.swift
dmhts: Done by accepting your suggestion.

giginet (approved):

PR #133, opened 2024-07-06
Support Swift 6 snapshot

giginet:
This PR is implemented based on `swift6` branch.

This PR replaces the default stack with Swift6 Snapshot while keeping Swift 5.10 compatibility.

### DIFF
--- a/Package@swift-5.10.swift
+++ b/Package@swift-5.10.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 6.0
+// swift-tools-version: 5.10
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
@@ -23,7 +23,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-package-manager.git",
-                 branch: "release/6.0"),
+                 revision: "swift-5.10-RELEASE"),
         .package(url: "https://github.com/apple/swift-log.git",
                  from: "1.5.2"),
         .package(url: "https://github.com/apple/swift-collections",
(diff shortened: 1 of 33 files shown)


Package@swift-5.10.swift
```diff
@@ -0,0 +1,89 @@
+// swift-tools-version: 5.10
```
giginet, line 1: This file is just copied from the current Package.swift

Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -182,19 +189,27 @@ extension DescriptionPackage {
         return products.reversed()
     }
 
-    private func targetsToBuild() throws -> Set<ResolvedTarget> {
+    private func targetsToBuild() throws -> [ScipioResolvedModule] {
```
giginet, line 192: Since Swift 6, `modules` returns `IdentifiableSet`, it's not available in the old SwiftPM, so it converts it to a regular Array.

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -77,6 +77,25 @@ extension PinsStore.PinState: Codable {
     }
 }
 
+#if compiler(>=6.0)
+
+extension PinsStore.PinState: @retroactive Hashable {
```
giginet, line 82: `@retroactive` is not supported on Swift 5, so these implementations are just copied.

Sources/scipio/CreateCommands.swift
```diff
@@ -1,11 +1,11 @@
 import Foundation
 import ScipioKit
-import ArgumentParser
+@preconcurrency import ArgumentParser
```
giginet, line 3: SwiftPM still depends on an old swift-argument-parser. It doesn't support concurrency, so we have to annotate `preconcurrency`

giuk-jung: Question: What happens if there is no annotate `preconcurrency`?

giuk-jung (approved):
LGTM

Sources/scipio/CreateCommands.swift
giginet: `CommandConfiguration` should be `Sendable`, but this version doesn't support it yet.

freddi-kit (approved):

PR #134, opened 2024-07-07
Improve `DescriptionPackage.resolveBuildProducts()` performance

Ryu0118:
## Motivation
In my personal application, the package contains a large number of targets, which causes the `resolveBuildProducts()` function to take a significant amount of time to execute. The common libraries added as dependencies to multiple targets lead to redundant generation of the same BuildProducts, resulting in overhead.

## Proposed Solution
To address this issue, I created a `BuildProductsResolver` class. This class caches the targets passed to the `buildProducts(from:)` function in a visitedTargets property. By doing this, when the same target is encountered again, it returns early, preventing redundant processing. This change significantly reduces execution time by eliminating repeated operations.

### DIFF
--- a/Sources/ScipioKit/DescriptionPackage.swift
+++ b/Sources/ScipioKit/DescriptionPackage.swift
@@ -150,108 +150,8 @@ struct DescriptionPackage {
 
 extension DescriptionPackage {
     func resolveBuildProducts() throws -> [BuildProduct] {
-        let targetsToBuild = try targetsToBuild()
-        var products = try targetsToBuild.flatMap(resolveBuildProduct(from:))
-
-        let productMap: [String: BuildProduct] = Dictionary(products.map { ($0.target.name, $0) }) { $1 }
-        func resolvedTargetToBuildProduct(_ target: ScipioResolvedModule) -> BuildProduct {
-            guard let product = productMap[target.name] else {
-                preconditionFailure("The dependency target (\(target.name)) was not found in the build target list")
-            }
-            return product
-        }
-
-        do {
-            products = try topologicalSort(products) { (product) in
-                return product.target.dependencies.flatMap { (dependency) -> [BuildProduct] in
-                    switch dependency {
-                    #if compiler(>=6.0)
-                    case .module(let module, conditions: _):
-                        return [resolvedTargetToBuildProduct(module)]
-                    case .product(let product, conditions: _):
-                        return product.modules.map(resolvedTargetToBuildProduct)
-                    #else
-                    case .target(let target, conditions: _):
-                        return [resolvedTargetToBuildProduct(target)]
-                    case .product(let product, conditions: _):
-                        return product.targets.map(resolvedTargetToBuildProduct)
-                    #endif
-                    }
-                }
-            }
-        } catch {
-            switch error {
-            case GraphError.unexpectedCycle: throw Error.cycleDetected
-            default: throw error
-            }
-        }
-
-        return products.reversed()
-    }
-
-    private func targetsToBuild() throws -> [ScipioResolvedModule] {
-        switch mode {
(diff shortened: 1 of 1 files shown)


Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -149,10 +149,58 @@ struct DescriptionPackage {
...
+        // build product must be the same regardless of the triple. Meanwhile, the target name and
+        // package identity remain relevant and unambiguously identify the build product.
+        hasher.combine(target.name)
+        hasher.combine(package.identity)
+    }
+}
+
+
+private final class BuildProductsResolver {
+    private var visitedTargets: Set<ResolvedTarget> = []
```
giginet, line 191: This may breaks since Swift 6 (https://github.com/giginet/Scipio/pull/133) because `ResolvedTarget` is no more conformed to `Hashable` since the new version.

Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -220,43 +245,39 @@ extension DescriptionPackage {
     }
 
     private func buildProducts(from target: ResolvedTarget) throws -> Set<BuildProduct> {
-        guard let package = graph.package(for: target) else {
+        guard let package = descriptionPackage.graph.package(for: target),
+              !visitedTargets.contains(target)
+        else {
             return []
         }
```
giginet, line 252: [Q] If the target was already visited, should it return just an empty set? Don't we need to keep the products resolved?

giginet (commented):
This basic concept seems to be fine. However, we're supporting It for Swift 6 snapshot.
These implementations may break when SwiftPM was updated.

Merge this PR before #133, or could you follow its changes after merging that PR?

Sources/ScipioKit/DescriptionPackage.swift
Ryu0118: Thank you for your feedback.
The early return of an empty set when a target has already been visited is intentional and should not cause issues in the current implementation.

The return value of `buildProducts()` is used in:

```Swift
Set(try rootTarget.recursiveTargetDependencies().flatMap(buildProducts(from:)))
```
in `resolveBuildProduct()`

```Swift
let productMap: [String: BuildProduct] = Dictionary(products.map { ($0.target.name, $0) }) { $1 }
```
in `resolveBuildProducts()`

In both cases, duplicate `BuildProduct` are eliminated. The early return of an empty set using `visitedTargets` is intended to **preemptively** remove part of the duplicates, so I believe it is not an issue.

Ryu0118: However, since the output of the `buildProducts(from:)` can change between the first and second execution with the same input, it must be used under the assumption that there is logic to eliminate such duplicates. Therefore, as you suggested, it might be better to keep the products resolved.

Ryu0118:
I will implement Swift 6 support on this branch after #133 is merged.

Ryu0118:
@giginet OK, I have fixed it. Could you review these changes again?

freddi-kit (commented):
implementation is reasonable for me 👍 
@Ryu0118 Could you show me the result of this refactoring? How times it speed up?

Ryu0118:
@freddi-kit I tested the execution of `resolveBuildProducts()` in [my OSS project](https://github.com/Ryu0118/TaskTree), and the time was reduced from 2.4 seconds in the previous implementation to 0.006 seconds, resulting in an approximately 392 times speed improvement.

giginet:
Sorry, the current main seems to be broken. I fixed this on #135. Please moment.

giginet (approved):
Great change! It makes sense to use memorization.

I can't wait to measure the impact of our product.
Thank you 😄

PR #136, opened 2024-07-26
Respect excluded files for modulemap generation

giginet:
ModuleMapGenerator generates modulemap for distributed frameworks.
However, currently, it ignores `excludes` settings on the package manifest.

This change make that it respects `excludes` settings of the package manifest to generate modulemap.

Currently, it's difficult to write test for `ModuleMapGenerator`, so this should be refactored.

### DIFF
--- a/Sources/ScipioKit/Producer/PIF/FrameworkModuleMapGenerator.swift
+++ b/Sources/ScipioKit/Producer/PIF/FrameworkModuleMapGenerator.swift
@@ -69,7 +69,7 @@ struct FrameworkModuleMapGenerator {
 
                 let targetDescription = descriptionPackage.targetDescription(of: clangTarget.name)
 
-                let includingHeaders = if let targetDescription {
+                let includingHeaders: Set<AbsolutePath> = if let targetDescription {
                     try Self.excludePaths(
                         from: allHeaders,
                         excludedFiles: Set(targetDescription.exclude.map { try RelativePath(validating: $0) }),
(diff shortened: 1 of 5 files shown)


Sources/ScipioKit/Producer/PIF/FrameworkModuleMapGenerator.swift
```diff
@@ -64,8 +65,20 @@ struct ModuleMapGenerator {
...
+                let includingHeaders: Set<AbsolutePath> = if let targetDescription {
+                    try Self.excludePaths(
+                        from: allHeaders,
+                        excludedFiles: Set(targetDescription.exclude.map { try RelativePath(validating: $0) }),
+                        targetRoot: clangTarget.path.scipioAbsolutePath
+                    )
+                } else {
+                    allHeaders
+                }
+                let declarations = includingHeaders.map { "    header \"\($0.basename)\"" }
```
giuk-jung, line 81: Just a question, is Tab necessary?

giginet: Yes, it will be parts of generated modulemap

```modulemap
framework module Foo {
    header "aaa.h"
}
```

giginet (commented):
Thank you for reviewing but I opened a new PR https://github.com/giginet/Scipio/pull/138/files

This PR's approach would not work for some situations because `TargetDescription` is not found. So I'll close this.

PR #140, opened 2024-09-09
Task / Configurable runner environment

dmhts:
### Motivation:
The motivation for this PR is to make `Runner` more flexible by optionally allowing the provision of a custom environment for the underlying `UserToolchain`. 

### Context:
In our company, we use ScipioKit in a quite advanced way. Specifically, we use it to precompile packages with multiple toolchains within a single run. For this, we set certain environment variables such as `PATH` and `DEVELOPER_DIR` before each Runner's run to delicately control the compiling process. 

It all works well, except for the case when the environment is not provided to `UserToolchain`; it [defaults to `Environment.current`](https://github.com/swiftlang/swift-package-manager/blob/b348d44d292070b14db56f317f3a4ba96305a8f9/Sources/PackageModel/UserToolchain.swift#L616). This is not a problem by itself; the problem is that such an environment is [statically cached](https://github.com/swiftlang/swift-package-manager/blob/b348d44d292070b14db56f317f3a4ba96305a8f9/Sources/Basics/Environment/Environment.swift#L121) when called for the very first time, and there's no way to redefine it in the subsequent runs (as all the Environment state modifiers are internal).

### Solution:
This problem can be easily fixed by extending the `Runner` interface so it takes an optional `Environment` parameter in its build options, which is then handed over to the underlying `UserToolchain`.

### Important: 
The PR ensures that the newly introduced environment property (in `BuildOptions`) does not affect the existing `SwiftPMCacheKey`, meaning the cache will remain environment-agnostic.

@giginet I'd love to hear your opinion on this.


Sources/scipio/CreateCommands.swift
```diff
@@ -52,12 +52,4 @@ extension Scipio {
...
 
-#if compiler(>=6.0)
-
-extension Runner.Options.Platform: @retroactive ExpressibleByArgument { }
-
-#else
-
 extension Runner.Options.Platform: ExpressibleByArgument { }
-
-#endif
```
dmhts, line 63: Screenshot 2024-09-09 at 17 04 54

These changes are not needed anymore since `Beta 6`.

Sources/scipio/Options.swift
```diff
@@ -49,7 +49,7 @@ struct BuildOptionGroup: ParsableArguments {
 
 #if compiler(>=6.0)
 
-extension FrameworkType: @retroactive ExpressibleByArgument { }
+extension FrameworkType: ExpressibleByArgument { }
```
dmhts, line 52: Same here.

Sources/ScipioKit/Runner.swift
```diff
@@ -175,7 +180,8 @@ extension Runner {
                 extraFlags: ExtraFlags? = nil,
                 extraBuildParameters: [String: String]? = nil,
                 enableLibraryEvolution: Bool? = nil,
-                frameworkModuleMapGenerationPolicy: FrameworkModuleMapGenerationPolicy? = nil
+                frameworkModuleMapGenerationPolicy: FrameworkModuleMapGenerationPolicy? = nil,
+                environment: [String: String]? = nil
```
dmhts, line 184: No changes for existing Scipio clients are required. As I said, it's rather a knob for advanced use cases like ours.

dmhts:
It seems the `Environment` API is not available in last year's SwiftPM. I'm going to make the respective changes.

Sources/ScipioKit/Runner.swift
```diff
@@ -154,6 +161,34 @@ extension Runner {
...
+                platforms: PlatformSpecifier = .manifest,
+                isSimulatorSupported: Bool = false,
+                isDebugSymbolsEmbedded: Bool = false,
+                frameworkType: FrameworkType = .dynamic,
+                extraFlags: ExtraFlags? = nil,
+                extraBuildParameters: [String: String]? = nil,
+                enableLibraryEvolution: Bool = false,
+                frameworkModuleMapGenerationPolicy: FrameworkModuleMapGenerationPolicy = .autoGenerated,
+                environment: Environment? = nil
+            ) {
```
dmhts, line 177: Alternatively, instead of passing the `Environment` type, we could pass a plain `[String: String]` dictionary containing the environment variables. This dictionary could later be mapped to the Environment type, allowing us to avoid the checks.

However, once support for `Xcode 15` is dropped, these checks will be removed anyway. Just tell me what you prefer.

Sources/ScipioKit/BuildOptions.swift
```diff
@@ -1,7 +1,20 @@
...
+    private enum CodingKeys: CodingKey {
+        case buildConfiguration
+        case isDebugSymbolsEmbedded
+        case frameworkType
+        case sdks
+        case extraFlags
+        case extraBuildParameters
+        case enableLibraryEvolution
+        case customFrameworkModuleMapContents
+    }
```
giginet, line 16: Can you tell me why did you add the coding keys?

We already use Codable without CodingKeys using `KeyEncodingStrategy`, so we're better to avoid adding `CodingKey` it was not required.

Sources/ScipioKit/BuildOptions.swift
```diff
@@ -33,6 +72,10 @@ struct BuildOptions: Hashable, Codable, Sendable {
     /// - Note: It have to store the actual file contents rather than its path,
     /// because the cache key should change when the file contents change.
     let customFrameworkModuleMapContents: Data?
+
+    #if compiler(>=6.0)
+    private(set) var environment: Environment?
```
giginet, line 77: How about renaming this like `swiftPMEnvironment`? I feel `Environment` seems to be too generic a name.

Sources/ScipioKit/Runner.swift
giginet: > However, once support for Xcode 15 is dropped, these checks will be removed anyway. 

Does this mean SwiftPM for Swift 5.x has a different interface?

I prefer to use a plane `[String: String]` in this case. Because `Runner` will be published to the external packages so we wish to keep fewer dependencies in the public interfaces as possible.
Using `Environment` here, we need to depend on SwiftPM from the caller side.

There is a problem with the keys don't become type-safe, but we may add a runtime check.

giginet (commented):
Sorry to keep you waiting!

I understood the requirements and wish to merge this
I commented some points.

Sources/ScipioKit/Runner.swift
dmhts: That's a good point, then I'll switch over to [String:String].

One thing, I'm on vacation till the end of the next week. Once back, I'll make the changes!

giginet: Never mind. Enjoy!

dmhts: Thanks! I'll answer to the other comments when I'm back.

Sources/scipio/CreateCommands.swift
giginet: These incompatibilities are already resolved on #141. Please rebase the PR.

Sources/ScipioKit/Runner.swift
dmhts: Hey, I'm finally back, and I've made the changes we discussed above.

Sources/ScipioKit/BuildOptions.swift
dmhts: I added `CodingKeys` to exclude the `environment` property from the encoding process. If you remove it the `testEncodeCacheKey()` test would fail as it would contain the environment in the output JSON:
```
"buildOptions" : {
    "buildConfiguration" : "release",
    "enableLibraryEvolution" : true,
    "environment" : {
      "DEVELOPER_DIR" : "\/Xcode.app\/Contents\/Developer",
      "PATH" : "\/Xcode.app\/Contents\/Developer:\/usr\/bin:\/bin"
    },
    ...
```

Sources/ScipioKit/BuildOptions.swift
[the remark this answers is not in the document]
dmhts: > Cannot automatically synthesize `Decodable` because `environment` does not have a matching CodingKey and does not have a default value.

The reason I use `var` in combination with a private setter is to avoid the error above. So to exclude `environment` from the `codable` process we have to:
- use the custom coding keys with a private setter or 
- implement a custom `Encodable/Decodable` conformance (which would be even more code).

Let me know if you see an alternative approach.

giginet: `toolchainEnvironment` seems to be a suitable name!

giginet (commented):
I'm sorry for asking so late.

I want to clarify whether changes to the Environment affect the cache.
In your current example, you are passing `DEVELOPER_DIR` and `PATH`. As such, these changes will not affect the final artifacts.

I'd like to know whether there are any other configuration values that could affect the final product due to changes in the Environment.

And if it is not necessary to incorporate Environment changes into CacheKey, it seems better to pass Environment directly to `Runner` rather than having it in `BuildOptions`.


`BuildOptions` contains configuration values that are directly included in CacheKey, so it may cause confusion if some values are not included depending on the property
How about a design where the values are passed to `Runner.Options` instead of `BuildOptions` and then propagated?

dmhts:
@giginet What you say sounds reasonable, it would lead to a better architecture indeed. Additionally, environment configuration should be done once per run, hence it fits more when done on the runner level.

- As pros, reasoning about build options is easier, requires no exceptions, and less code overall. 
- As a con, I had to extend both `FrameworkProducer` and `PIFCompiler` initializers as we don't pass the runner options directly there. I believe it's a very minor change though, and it still remains an implementation detail.

I just applied the changes as suggested [here](https://github.com/giginet/Scipio/pull/140/commits/ef8fb172b21d762ba8191d8275c9d6d2d8193b0d).

dmhts:
I didn't add any tests since there are no existing test cases for `PIFCompiler` or `ToolchainGenerator`. I believe integration tests are more appropriate for covering this change, as it primarily concerns configuration passed down to SwiftPM.

Sources/ScipioKit/Producer/PIF/PIFCompiler.swift
```diff
@@ -18,27 +19,29 @@ struct PIFCompiler: Compiler {
...
+        self.toolchainEnvironment = toolchainEnvironment
     }
 
     private func fetchDefaultToolchainBinPath() async throws -> AbsolutePath {
         let result = try await executor.execute("/usr/bin/xcrun", "xcode-select", "-p")
-        let rawString = try result.unwrapOutput()
+        let rawString = try result.unwrapOutput().trimmingCharacters(in: .whitespacesAndNewlines)
         let developerDirPath = try AbsolutePath(validating: rawString)
         let toolchainPath = try RelativePath(validating: "./Toolchains/XcodeDefault.xctoolchain/usr/bin")
         return developerDirPath.appending(toolchainPath)
```
dmhts, line 39: @giginet some context: without this change, the `toolchainPath` was interpolated as `/Applicaitons/Xcode.app/Developer/Contents/.\nToolchains/XcodeDefault.xctoolchain/usr/bin`. It has worked so far because SwiftPM down the road defaulted to the correct toolchain path, in the end. However, when run with different environments within the same runner it led to incorrect toolchain path determination. 

I added this change ad-hoc but it might make sense to trim whitespaces and new lines inside of the `unwrapOutput()` function so potential issues are also addressed in other places. I can do it withing this PR or you can just take it FYI.

giginet: This change will not affect other implementations. Thank you.

giginet (commented):
Almost looks great!

Thank you for the quick fixes.
I pointed trivial points.

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -51,7 +52,8 @@ struct FrameworkProducer {
         cacheMode: Runner.Options.CacheMode,
         overwrite: Bool,
         outputDir: URL,
-        fileSystem: any FileSystem = localFileSystem
+        fileSystem: any FileSystem = localFileSystem,
+        toolchainEnvironment: [String: String]? = nil
```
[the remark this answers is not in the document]
dmhts: Of course, I'll be following this convention in the future.

giginet (approved):

giginet:
@dmhts Thank you for the contribution! I'll release the new version soon.

By the way, I'm very excited that your company uses this product heavily. I'm happy to hear the case study someday.

dmhts:
@giginet That's not it, there's one more PR coming soon :) 

We're currently working on a tool that uses SwiftPM and Scipio at its core. Once everything is polished, our goal is to open-source it. It's already running successfully in our CI pipelines. You'll be one of the first people I let know when it's ready!

PR #141, opened 2024-09-12
Fix build errors in Xcode16-RC

simorgh3196:
I have fixed an error that occurs when building with Xcode 16 Release Candidate.
This change is also included in #140, but I wanted to use it immediately, so I created a small separate PR.

### Changes

* Removed unnecessary `@retroactive` declarations

### DIFF
--- a/Sources/scipio/Arguments.swift
+++ b/Sources/scipio/Arguments.swift
@@ -10,19 +10,6 @@ extension URL: @retroactive ExpressibleByArgument {
     }
 }
 
-extension BuildConfiguration: @retroactive ExpressibleByArgument {
-    public init?(argument: String) {
-        switch argument.lowercased() {
-        case "debug":
-            self = .debug
-        case "release":
-            self = .release
-        default:
-            return nil
-        }
-    }
-}
-
 #else
 
 extension URL: ExpressibleByArgument {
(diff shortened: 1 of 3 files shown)


Sources/scipio/Arguments.swift
```diff
@@ -10,19 +10,6 @@ extension URL: @retroactive ExpressibleByArgument {
...
-        case "debug":
-            self = .debug
-        case "release":
-            self = .release
-        default:
-            return nil
-        }
-    }
-}
-
```
giginet, line 25: This extension was also required for Swift 5 & 6.

Please move L34 to the outside of the preprocessor block.

giginet (commented):
Thank you!

Could you fix the commented points?

Sources/scipio/Arguments.swift
simorgh3196: https://github.com/giginet/Scipio/blob/c062eeb00bfd62618a7e316cab725a318262c77b/Sources/scipio/Arguments.swift

The `BuildConfiguration` extension is already outside the preprocessor block.
If my perception is wrong, I would like to know!

giginet: oops. It's correct!

giginet (approved):
LGTM!

These same changes are applied on #140, but we should merge this PR first.
I'll ask the author to rebase this PR.

PR #142, opened 2024-09-14
Add x86_64 binary to an artifact bundle in a release

mtj0928:
I noticed that we can add x86_64 binary to an artifact bundle by build the project twice.

When scipio is built for arm64, the binary is generated at `./.build/arm64-apple-macosx/release/scipio`.
And when it is built for x86_64, the binary is generated at `./.build/x86_64-apple-macosx/release/scipio`.

[ArtifactBundleGen](https://github.com/freddi-kit/ArtifactBundleGen) used by scipio can combines both binaries into an artifact bundle when they exist.

This is the artifact bundle that I made in my forked repository.
https://github.com/mtj0928/Scipio/releases/download/0.21.0/scipio.artifactbundle.zip

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,8 +18,8 @@ jobs:
           swift package resolve
       - name: Build Binary
         run: |
-          # Scipio can't build for x86_64 on Apple Silicon because it uses build tools plugin
           swift build --disable-sandbox -c release --arch arm64
+          swift build --disable-sandbox -c release --arch x86_64
       - name: Get Current Tag
         run: echo "TAG_NAME=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
       - name: Make Artifact Bundle


mtj0928:
Oh, a new rule, `optional_data_string_conversion`, was added to SwiftLint in [0.57.0](https://github.com/realm/SwiftLint/releases/tag/0.57.0), which was released a few days ago.
It seems that the test code in Scipio violates this rule.

giginet (commented):
I didn't know the update! It may be introduced since SwiftPM 6.0.


On CI, we are still using Xcode 15.4. It may use `Package@swift-5.10`. This job was only executed on PR was merged into the main so `x86_64` support would not be worked well.

```
DEVELOPER_DIR: "/Applications/Xcode_15.4.app/Contents/Developer"
```

Could you update DEVELOPER_DIR to Xcode 16?

mtj0928:
Actually, this is not Swift 6 feature!
I could success the job in my forked repository, and the job also used Xcode 15.4!
https://github.com/mtj0928/Scipio/actions/runs/10858648433/job/30136902542

giginet (approved):

PR #146, opened 2024-10-01
Task / Enable root package target caching

dmhts:
### Motivation:
Currently, `Scipio` does not support caching root package (local, if you will) targets. Apparently, this is because the version requirement is typically known for the root package dependencies, but not for the root package's direct targets themselves (as they're not pinned in the underlying `PinStore`, for obvious reasons). However, such a version requirement can be easily derived from the git repository that "wraps" the root package, and this is what this PR is about. Moreover, SwiftPM requires each package to be within a repository with the respective version requirement set.

### Context:
We're using `Scipio` only with the `.createPackage` mode where the runner goes over a list of 3rd party packages. The issue is reproducible specifically in that mode, and you also can see it happening in the integration tests if run in the `.createPackage` mode with the caching flag enabled; the reason is that fixture packages are not wrapped in git repositories (which normally shouldn't happen in real life), otherwise, this PR fixes a potential issue there too.

### Additional considerations:
The change is purely additive and doesn't alter any existing functionality.

###  Expected result:
Before: 
🚀 Cache MyTarget.xcframework to cache storage
⚠️ Can't create caches for .../TestingPackage/MyTarget.xcframework

After: 
🚀 Cache MyTarget.xcframework to cache storage
[No warning as cache created successfuly]

@giginet as always, I'd love to hear your thoughts on this.


Tests/ScipioKitTests/CacheSystemTests.swift
```diff
@@ -1,31 +1,22 @@
...
 
-    func testParseClangVersion() async throws {
-        let hook = { arguments in
-            XCTAssertEqual(arguments, ["/usr/bin/xcrun", "clang", "--version"])
-            return StubbableExecutorResult(arguments: arguments, success: self.clangVersion)
-        }
-        let clangParser = ClangChecker(executor: StubbableExecutor(executeHook: hook))
-        let version = try await clangParser.fetchClangVersion()
-        XCTAssertEqual(version, "clang-1400.0.29.102")
-    }
```
dmhts, line 27: I hope it's fine that I extracted the Clang version checking logic into a separate test case.

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -306,3 +312,32 @@ public struct VersionFileDecoder {
         )
     }
 }
+
+#if compiler(>=6.0)
+
+fileprivate extension ResolvedPackage {
+
+    var pin: PinsStore.Pin? {
```
giginet, line 320: I feel `pin` is too generic a property name, even if it's a fileprivate extension.

This property generates a pin it is using `GitRepository`

How about it be `makePinByRevision()` or something?

Tests/ScipioKitTests/CacheSystemTests.swift
```diff
@@ -45,38 +36,112 @@ InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault
...
+
+        // Ensure that the cache key cannot be calculated if the package is not in the Git repository.
+        do {
+            _ = try await cacheSystem.calculateCacheKey(of: cacheTarget)
+            XCTFail("A cache key should not be possible to calculate if the package is not in a repository.")
+        } catch let error as CacheSystem.Error {
+            XCTAssertEqual(error.errorDescription, "Repository version is not detected for \(descriptionPackage.name).")
+        } catch {
+            XCTFail("Wrong error type.")
+        }
```
giginet, line 130: It's better to use `XCTAssertThrowErrors` instead
(Sorry for the other code uses this)

giginet (commented):
Thank you for the contribution!

Your concept looks good. I commented to minor points.

Tests/ScipioKitTests/CacheSystemTests.swift
dmhts: Since `calculateCacheKe`y is an async function and there is no version of `XCTAssertThrowErrors` that supports concurrency, using the latter with the former becomes problematic. Let me know if you still see how it can be applied to this context.

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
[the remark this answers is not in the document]
dmhts: Done.

dmhts: Agree, I like using an optional `makePinFromRevision()` method for that. I just slightly changed the preposition.

Tests/ScipioKitTests/CacheSystemTests.swift
giginet: Make sense! Sorry for my wrong point

giginet (approved):

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -269,14 +271,18 @@ struct CacheSystem: Sendable {
         )
     }
 
-    private func retrievePin(product: BuildProduct) throws -> PinsStore.Pin {
-        #if swift(>=5.10)
-        guard let pin = pinsStore.pins[product.package.identity] else {
-            throw Error.revisionNotDetected(product.package.manifest.displayName)
+    private func retrievePin(package: ResolvedPackage) throws -> PinsStore.Pin {
+        #if compiler(>=6.0)
+        guard let pin = pinsStore.pins[package.identity] ?? package.makePinFromRevision() else {
```
ikesyo, line 276: I'm not sure why the addition is gated by `#if compiler(>=6.0)`. Is that change not compatible with Swift 5.10 compiler?

ikesyo (approved):

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
ikesyo: I understand that `GitRepository.getCurrentTag` is added in Swift 6.0 👍

dmhts: Thanks for the question. I should've mentioned that in the PR description. There are two reasons for this: `GitRepository.getCurrentTag` is only available in SwiftPM's `release/6.0` branch. Reimplementing the function would, of course, be trivial, but as I understand it, support for `Xcode 15` is going to be dropped soon, so I thought it wouldn't be worth the effort. Correct me if I'm wrong.

giginet: Good point. it is correct it will cause the incompatibility.

But I told to @dmhts  that we don't need to maintain the backward compatibility.
https://github.com/giginet/Scipio/pull/140#discussion_r1753282276

So, I think it's no problem if the compatibility is broken because Swift 5 support will be dropped soon. We don't need to make an effort to keep backward compatibility.

I'll drop the Swift 5 support.

PR #149, opened 2024-10-09
Update actions/checkout to v4

ikesyo:
To address node runtime deprecations:
- https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
- https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

### DIFF
--- a/.github/workflows/docc.yml
+++ b/.github/workflows/docc.yml
@@ -10,7 +10,7 @@ jobs:
   DocC:
     runs-on: macos-14
     steps:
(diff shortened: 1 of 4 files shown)


ikesyo:
We'd like to introduce Dependabot for updating such dependencies.

giginet (approved):

PR #150, opened 2024-10-17
Drop Swift 5 support

giginet:
This PR drops Swift 5 support.

Since this, we have to use Xcode 16 to develop, but it can be used by Swift 5 code.


Plugins/GenerateScipioVersion/plugin.swift
```diff
@@ -68,9 +60,6 @@ struct GenerateScipioVersion: BuildToolPlugin {
     }
 }
 
-#if compiler(<6.0)
-
-// Backward compatibility below 6.0 compiler
 // Convert Foundation.URL to Path
 extension Command {
     fileprivate static func prebuildCommand(
```
ikesyo, line 65: Why is this not removed?

giginet: Nice catch. My change was wrong

ikesyo (approved):

PR #151, opened 2024-10-22
Factor out DescriptionPackage's path-related functions to PackageLocator protocol

ikesyo:
This would make it easy for testing FrameworkComponentsCollector and FrameworkModuleMapGenerator.

### DIFF
--- a/Sources/ScipioKit/Producer/PIF/XCBuildClient.swift
+++ b/Sources/ScipioKit/Producer/PIF/XCBuildClient.swift
@@ -4,25 +4,25 @@ import PackageGraph
 import PackageModel
 
 struct XCBuildClient {
-    private let descriptionPackage: DescriptionPackage
     private let buildOptions: BuildOptions
     private let buildProduct: BuildProduct
     private let configuration: BuildConfiguration
+    private let packageLocator: any PackageLocator
     private let fileSystem: any FileSystem
     private let executor: any Executor
 
     init(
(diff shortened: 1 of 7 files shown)


Sources/ScipioKit/Producer/PIF/XCBuildClient.swift
```diff
@@ -4,25 +4,25 @@ import PackageGraph
...
     private let executor: any Executor
 
     init(
-        package: DescriptionPackage,
         buildProduct: BuildProduct,
         buildOptions: BuildOptions,
         configuration: BuildConfiguration,
+        packageLocator: some PackageLocator,
         fileSystem: any FileSystem = localFileSystem,
         executor: any Executor = ProcessExecutor(decoder: StandardOutputDecoder())
```
giginet, line 20: They also should be `some`

giginet (approved):

Sources/ScipioKit/Producer/PIF/XCBuildClient.swift
ikesyo: For `localFileSystem`, that is TSCBasic's global variable and that is typed as `any FileSystem` so cannot be changed. I'll fix `executor` part.

ikesyo: Addressed in 67a3cad

ikesyo:
🚢

PR #156, opened 2024-11-08
Add multiple cache storages feature

### LINKED ISSUE #148: Support cache redundancy (cache hierarchy)
Currently Scipio only supports one cache storage for `storage` cache mode. But there would be a demand for having two (or more) cache storages like: one as a primary storage, and another as a secondary (fallback) storage. For example,

- On CI machine (frameworks baking job): primary will be remote (S3) storage with producer-only and no secondary
- On developer machine: primary will be remote (S3) storage with consumer-only and secondary will be local storage with both producer and consumer

refs:
- https://github.com/giginet/Scipio/blob/2b6d90255384aae0f371e5b5f93098548b167c47/Sources/ScipioKit/Runner.swift#L205-L216
- https://github.com/mtj0928/scipio-s3/blob/336fd33acda725641197cedf00f49eb29fc67922/Sources/ScipioS3/CompositeCacheStorage.swift

ikesyo:
Resolves #148.

- `Runner.Options.CacheMode` is replaced with `[Runner.Options.CachePolicy]`
  - `cacheMode: .disabled` -> `cachePolicies: []`
  - `cacheMode: .project` -> `cachePolicies: [.project]`
  - `cacheMode: .storage(storage, actors)` -> `cachePolicies: [.init(storage: storage, actors: actors)]`
- When saving, cache frameworks into all storages with `CacheActorKind.producer`
- When restoring, storages with `CacheActorKind.consumer` are used in order
  - If first storage failed to restore some caches, the failed ones are tried to be fetched from second (next) storage

### DIFF
--- a/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
+++ b/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
@@ -108,7 +108,7 @@ struct CacheSystem: Sendable {
     static let defaultParalellNumber = 8
     private let pinsStore: PinsStore
     private let outputDirectory: URL
-    private let storage: (any CacheStorage)?
+    private let writableStorages: [any CacheStorage]
     private let fileSystem: any FileSystem
 
     struct CacheTarget: Hashable, Sendable {
@@ -139,17 +139,29 @@ struct CacheSystem: Sendable {
     init(
         pinsStore: PinsStore,
         outputDirectory: URL,
-        storage: (any CacheStorage)?,
+        writableStorages: [any CacheStorage],
         fileSystem: any FileSystem = localFileSystem
     ) {
         self.pinsStore = pinsStore
         self.outputDirectory = outputDirectory
-        self.storage = storage
+        self.writableStorages = writableStorages
         self.fileSystem = fileSystem
     }
 
     func cacheFrameworks(_ targets: Set<CacheTarget>) async {
-        let chunked = targets.chunks(ofCount: storage?.parallelNumber ?? CacheSystem.defaultParalellNumber)
+        guard !writableStorages.isEmpty else {
+            // About `CacheMode.project` which does not have any writableStorages, we don't need to do anything.
+            // The built frameworks under the project themselves are treated as valid caches.
+            return
+        }
+
+        for storage in writableStorages {
+            await cacheFrameworks(targets, storage: storage)
+        }
+    }
+
+    private func cacheFrameworks(_ targets: Set<CacheTarget>, storage: any CacheStorage) async {
+        let chunked = targets.chunks(ofCount: storage.parallelNumber ?? CacheSystem.defaultParalellNumber)
 
         for chunk in chunked {
             await withTaskGroup(of: Void.self) { group in
@@ -159,10 +171,10 @@ struct CacheSystem: Sendable {
                         let frameworkPath = outputDirectory.appendingPathComponent(frameworkName)
                         do {
                             logger.info(
-                                "🚀 Cache \(frameworkName) to cache storage",
+                                "🚀 Cache \(frameworkName) to cache storage: \(storage)",
                                 metadata: .color(.green)
                             )
-                            try await cacheFramework(target, at: frameworkPath)
+                            try await cacheFramework(target, at: frameworkPath, storage: storage)
                         } catch {
                             logger.warning("⚠️ Can't create caches for \(frameworkPath.path)")
                         }
@@ -173,10 +185,10 @@ struct CacheSystem: Sendable {
         }
     }
 
-    private func cacheFramework(_ target: CacheTarget, at frameworkPath: URL) async throws {
+    private func cacheFramework(_ target: CacheTarget, at frameworkPath: URL, storage: any CacheStorage) async throws {
         let cacheKey = try await calculateCacheKey(of: target)
 
-        try await storage?.cacheFramework(frameworkPath, for: cacheKey)
+        try await storage.cacheFramework(frameworkPath, for: cacheKey)
     }
 
     func generateVersionFile(for target: CacheTarget) async throws {
@@ -210,8 +222,8 @@ struct CacheSystem: Sendable {
         case failed(LocalizedError?)
         case noCache
     }
-    func restoreCacheIfPossible(target: CacheTarget) async -> RestoreResult {
-        guard let storage = storage else { return .noCache }
+
+    func restoreCacheIfPossible(target: CacheTarget, storage: any CacheStorage) async -> RestoreResult {
         do {
             let cacheKey = try await calculateCacheKey(of: target)
             if try await storage.existsValidCache(for: cacheKey) {
@@ -223,12 +235,6 @@ struct CacheSystem: Sendable {
         } catch {
             return .failed(error as? LocalizedError)
         }
-    }
-
-    private func fetchArtifacts(target: CacheTarget, to destination: URL) async throws {
-        guard let storage = storage else { return }
-        let cacheKey = try await calculateCacheKey(of: target)
-        try await storage.fetchArtifacts(for: cacheKey, to: destination)
     }
 
     func calculateCacheKey(of target: CacheTarget) async throws -> SwiftPMCacheKey {
--- a/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
+++ b/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
@@ -145,13 +145,7 @@ struct CacheSystem: Sendable {
         self.fileSystem = fileSystem
     }
 
-    func cacheFrameworks(_ targets: Set<CacheTarget>, storages: [any CacheStorage]?) async {
-        guard let storages, !storages.isEmpty else {
-            // About `CacheMode.project` which is not tied to any (external) storages, we don't need to do anything.
-            // The built frameworks under the project themselves are treated as valid caches.
-            return
-        }
-
+    func cacheFrameworks(_ targets: Set<CacheTarget>, storages: [any CacheStorage]) async {
         for storage in storages {
             await cacheFrameworks(targets, storage: storage)
         }
--- a/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
+++ b/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
@@ -245,7 +245,10 @@ struct CacheSystem: Sendable {
             buildOptions: buildOptions,
             clangVersion: clangVersion,
             xcodeVersion: xcodeVersion,
-            scipioVersion: "578ce98d236e79dad3e473cb11153e867be07174" // TODO: revert
+            // Making the cache key compatible with 0.24.0 temporarily for easier debugging.
+            //
+            // TODO: revert this before merging
+            scipioVersion: "578ce98d236e79dad3e473cb11153e867be07174"
         )
     }
 
--- a/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
+++ b/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
@@ -154,6 +154,7 @@ struct CacheSystem: Sendable {
     private func cacheFrameworks(_ targets: Set<CacheTarget>, storage: any CacheStorage) async {
         let chunked = targets.chunks(ofCount: storage.parallelNumber ?? CacheSystem.defaultParalellNumber)
 
+        let storageType = type(of: storage)
         for chunk in chunked {
             await withTaskGroup(of: Void.self) { group in
                 for target in chunk {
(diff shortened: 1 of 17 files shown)


Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -108,7 +108,6 @@ struct CacheSystem: Sendable {
     static let defaultParalellNumber = 8
     private let pinsStore: PinsStore
     private let outputDirectory: URL
-    private let storage: (any CacheStorage)?
```
ikesyo, line 111: Due to the introduction of `.storages` cache mode, storages may be different on when saving caches and when restoring caches. So removed the property and passes by arguments instead.

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -141,22 +120,93 @@ struct FrameworkProducer {
...
+        case .disabled:
+            return []
+        case .project:
+            // For `.project`, just checking whether the valid caches (already built frameworks under the project)
+            // exist or not (not restoring anything from external locations).
+            return await restoreCachesForTargets(
+                availableTargets,
+                cacheSystem: cacheSystem,
+                cacheStorage: nil
+            )
```
ikesyo, line 139: here: https://github.com/giginet/Scipio/blob/f6cbef149f247b1a09d9966ead372625b298ba59/Sources/ScipioKit/Producer/FrameworkProducer.swift#L245-L259

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -141,22 +120,93 @@ struct FrameworkProducer {
...
+
+            logger.info(
+                "⏸️ Restoration finished with cache storage: \(logSuffix)",
+                metadata: .color(.green)
+            )
+
+            remainingTargets.subtract(restoredPerStorage)
+            if remainingTargets.isEmpty {
+                break
+            }
```
ikesyo, line 185: if all frameworks are successfully restored, we don't need to proceed to next cache storage.

Sources/scipio/scipio.docc/build-pipeline.md
```diff
@@ -253,3 +253,30 @@ You can specify it by a second argument of `.storage` cache mode.
...
+let runner = Runner(
+    mode: .prepareDependencies,
+    options: .init(
+        baseBuildOptions: .init(
+            buildConfiguration: .release,
+            isSimulatorSupported: true
+        ),
+        cacheMode: .storages([
+            (s3Storage, [.consumer] as Set),
+            (localCacheStorage, [.producer, .consumer] as Set),
```
giginet, line 275: Do we need `as Set` explicitly? I think it would be influenced by `ArrayLiteralConvertible`

Sources/ScipioKit/Runner.swift
```diff
@@ -219,6 +219,7 @@ extension Runner {
             case disabled
             case project
             case storage(any CacheStorage, Set<CacheActorKind>)
```
giginet, line 221: How about using `static let` instead of case to reduce duplicated cases?

```swift
enum CacheMode {
  static let disabled: Self = .storages([])
  static func storage(storage: any CacheStorage, actor: Set<CacheActorKind>) -> Self {
      .storages([(storage, actor)])
  }
}
```

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -141,22 +120,93 @@ struct FrameworkProducer {
...
         availableTargets: Set<CacheSystem.CacheTarget>,
         cacheSystem: CacheSystem
     ) async -> Set<CacheSystem.CacheTarget> {
-        let chunked = availableTargets.chunks(ofCount: cacheStorage?.parallelNumber ?? CacheSystem.defaultParalellNumber)
+        let cacheStorages: [any CacheStorage]
+
+        switch cacheMode {
+        case .disabled:
+            return []
+        case .project:
```
giginet, line 132: It's very confusing `nil` means project cache.

How about adding `ProjectCacheStorage` to represent project cache mode?

By adding this, we can represent `.project` as `.storages([projectCacheStorage, [consumer, producer]])`

Sources/ScipioKit/Runner.swift
giginet, line 221: How about introducing a new struct to hold `any CacheStorage` and `Set<CacheActorKind>`?

```swift
struct CachePolicy {
    let cacheStorage: any CacheStorage
    let actorKinds: Set<CacheActorKind>
}
```

Sources/scipio/scipio.docc/build-pipeline.md
ikesyo: Unfortunately compilations fails without this somehow 😢

Sources/ScipioKit/Runner.swift
ikesyo: Yeah I agree, handling tuples is not a good experience 👍

ikesyo: I'll try it.

Sources/scipio/scipio.docc/build-pipeline.md
```diff
@@ -246,10 +248,37 @@ You can also implement your custom cache storage by implementing `CacheStorage`
...
+
+### Use multiple cache policies at the same time
+
+You can also use multiple cache policies, which accepts multiple cache storages with different sets of cache actors:
+
+```swift
+import ScipioS3Storage
+
+let s3Storage: some CacheStorage = ScipioS3Storage.S3Storage(config: ...)
+let localCacheStorage: some CacheStorage = LocalCacheStorage()
```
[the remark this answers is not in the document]
ikesyo: LocalDiskCacheStorage is public type and can be initialized with custom cache directory.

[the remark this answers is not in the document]
giginet: Yes!

Because we want to restrict their actors. e.g. for `ProjectCache`, users can only use `[.producer]` combination.

To conceal the storage constructed, we can prevent mistakes in usage.

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -303,6 +339,13 @@ struct FrameworkProducer {
         return []
     }
 
+    private func cacheFrameworksIfNeeded(_ targets: Set<CacheSystem.CacheTarget>, cacheSystem: CacheSystem) async {
+        let storagesWithProducer = cachePolicies.storages(for: .producer)
+        if !storagesWithProducer.isEmpty {
+            await cacheSystem.cacheFrameworks(targets, to: storagesWithProducer)
+        }
```
giginet, line 346: [Q] I'm concerned that the way storages are handled is different for consumers and producers.

For consumers, storages are evaluated in order and the top ones are used.
On the other hand, producers try to write to all declared storages at the same time.

Is this difference in behavior intentional?

ikesyo: It's intentional. If we don't cache built frameworks to 2nd storages and later, the storage will not be able to restore the frameworks so saving to all storages will be required I think.

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -247,7 +245,10 @@ struct CacheSystem: Sendable {
             buildOptions: buildOptions,
             clangVersion: clangVersion,
             xcodeVersion: xcodeVersion,
-            scipioVersion: currentScipioVersion
+            // Making the cache key compatible with 0.24.0 temporarily for easier debugging.
+            //
+            // TODO: revert this before merging
+            scipioVersion: "578ce98d236e79dad3e473cb11153e867be07174"
```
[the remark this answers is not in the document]
ikesyo: Disabling when developing is not a solution, because excluding `scipioVersion` from cache keys, those are not compatible with the cache keys generated by release-versioned Scipio.

Ideally we should introduce some Scipio's cache versioning and increment it when there are some incompatible changes to built frameworks (but it'll make Scipio's development and review process a bit complicated).

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -139,18 +138,23 @@ struct CacheSystem: Sendable {
...
 
-    func cacheFrameworks(_ targets: Set<CacheTarget>) async {
-        let chunked = targets.chunks(ofCount: storage?.parallelNumber ?? CacheSystem.defaultParalellNumber)
+    func cacheFrameworks(_ targets: Set<CacheTarget>, to storages: [any CacheStorage]) async {
+        for storage in storages {
+            await cacheFrameworks(targets, to: storage)
+        }
+    }
 
+    private func cacheFrameworks(_ targets: Set<CacheTarget>, to storage: any CacheStorage) async {
```
giginet, line 154: [nits] 
```suggestion
    private func cacheFrameworks(_ targets: Set<CacheTarget>, to storage: some CacheStorage) async {
```

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -210,8 +214,8 @@ struct CacheSystem: Sendable {
         case failed(LocalizedError?)
         case noCache
     }
-    func restoreCacheIfPossible(target: CacheTarget) async -> RestoreResult {
-        guard let storage = storage else { return .noCache }
+
+    func restoreCacheIfPossible(target: CacheTarget, storage: any CacheStorage) async -> RestoreResult {
```
giginet, line 218: ditto

```suggestion
    func restoreCacheIfPossible(target: CacheTarget, storage: some CacheStorage) async -> RestoreResult {
```

Sources/ScipioKit/Producer/FrameworkProducer.swift
giginet: [not mandatory] it's kindly to comment this

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -194,11 +189,58 @@ struct FrameworkProducer {
...
+            )
+            restored.formUnion(restoredPerStorage)
+
+            logger.info(
+                "⏸️ Restoration finished with cache storage: \(logSuffix)",
+                metadata: .color(.green)
+            )
+
+            remainingTargets.subtract(restoredPerStorage)
+            if remainingTargets.isEmpty {
```
giginet, line 229: it's kindly to comment this

```suggestion
            // if all frameworks are successfully restored, we don't need to proceed to next cache storage.
            if remainingTargets.isEmpty {
```

giginet (approved):
Almost great!

Sources/ScipioKit/Runner.swift
```diff
@@ -224,15 +224,26 @@ extension Runner {
...
                 actors: [.producer]
             )
 
-            public static let localDisk = Self(
-                storage: LocalDiskCacheStorage(),
+            /// The cache policy for saving to and restoring from the system cache directory `~/Library/Caches/Scipio`.
+            public static let localDisk: Self = Self(
+                storage: LocalDiskCacheStorage(baseURL: nil),
                 actors: [.producer, .consumer]
             )
```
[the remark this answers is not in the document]
ikesyo: To do so we need to change

```swift
- public static func localDisk(baseURL: URL)
+ public static func localDisk(baseURL: URL?)
```

In that case, I'd like to remove `static let localDisk` and make use of the default case as `cachePolicies: [.localDisk()]`

Sources/ScipioKit/Producer/Cache/LocalDiskCacheStorage.swift
```diff
@@ -3,37 +3,33 @@ import ScipioStorage
...
     enum Error: Swift.Error {
         case cacheDirectoryIsNotFound
     }
 
-    public enum CacheDirectory: Sendable {
-        case system
-        case custom(URL)
-    }
-
-    private let cacheDirectroy: CacheDirectory
```
giginet, line 20: :eyes: Directroy

Sources/ScipioKit/Runner.swift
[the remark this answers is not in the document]
ikesyo: To reuse `static func localDisk(baseURL: ...)` the `baseURL` argument must be optional. That means that if we have both

```swift
static let localDisk: Self
static func localDisk(baseURL: URL? = nil) -> Self
```

There is 2 or 3 ways of using the localDisk cache policy with the default system cache dir ( `.localDisk`, `.localDisk()` and `.localDisk(baseURL: nil)`). That is a bit confusing so I prefer either of

1. Define only one `static func localDisk(baseURL: URL? = nil) -> Self`
2. Define `static let localDisk: Self` and `static func localDisk(baseURL: URL) -> Self`

Is which one preferable for you?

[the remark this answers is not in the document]
ikesyo:
```swift
public static let localDisk: Self = localDisk(baseURL: nil)

public static func localDisk(baseURL: URL) -> Self {
    localDisk(baseURL: baseURL)
}

private static func localDisk(baseURL: URL?) -> Self {
    Self(
        storage: LocalDiskCacheStorage(baseURL: baseURL),
        actors: [.producer, .consumer]
    )
}
```

is also an option.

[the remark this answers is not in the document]
giginet: OK, let's reduce the redundance.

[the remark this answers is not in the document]
ikesyo: Addressed in bced0e2 👍

giginet (approved):

giginet (approved):

PR #158, opened 2024-11-12
Factor out the validation logic of existing frameworks in the output directory from Restorer

ikesyo:
That is not the actual restoration process, so should be done outside Restorer.

---

This is a preparation for #156.


giginet (approved):
🚀

PR #161, opened 2024-11-12
Update README.md

mackoj:
Add an installation section


giginet (approved):
Thank you! I'll add the information after merging!

PR #166, opened 2024-12-11
Remove ScipioAbsolutePath and use TSCAbsolutePath typealias instead

ikesyo:
- https://github.com/swiftlang/swift-package-manager/blob/bc689418acad746af6f2e24e6bef640e411d9bd7/Sources/Basics/FileSystem/AbsolutePath.swift#L16-L17
- https://github.com/swiftlang/swift-package-manager/blob/bc689418acad746af6f2e24e6bef640e411d9bd7/Sources/Basics/FileSystem/AbsolutePath.swift#L45-L48


ikesyo:
> We aim to reduce the usage of TSC in the future.

Yes, making the TSC types explicit should be  useful migration.

PR #168, opened 2024-12-12
Add canonical package location to SwiftPMCacheKey to differentiate the same target when its package is changed from remote one to local one (or vice-versa)

ikesyo:
Currently there is no package information in SwiftPMCacheKey.

ref: https://github.com/swiftlang/swift-package-manager/blob/bc689418acad746af6f2e24e6bef640e411d9bd7/Sources/PackageModel/PackageIdentity.swift#L16-L19

### DIFF
--- a/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
+++ b/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
@@ -96,6 +96,7 @@ extension PinsStore.PinState: @retroactive Hashable {
 }
 
 public struct SwiftPMCacheKey: CacheKey {
+    public var packageIdentity: PackageIdentity
     public var targetName: String
     public var pin: PinsStore.PinState
     var buildOptions: BuildOptions
@@ -230,6 +231,7 @@ struct CacheSystem: Sendable {
     }
 
     func calculateCacheKey(of target: CacheTarget) async throws -> SwiftPMCacheKey {
+        let packageIdentity = target.buildProduct.package.identity
         let targetName = target.buildProduct.target.name
         let pin = try retrievePin(package: target.buildProduct.package)
         let buildOptions = target.buildOptions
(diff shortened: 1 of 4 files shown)


freddi-kit (approved):

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -96,6 +96,7 @@ extension PinsStore.PinState: @retroactive Hashable {
 }
 
 public struct SwiftPMCacheKey: CacheKey {
+    public var packageIdentity: PackageIdentity
```
ikesyo, line 99: ref: https://github.com/swiftlang/swift-package-manager/blob/bc689418acad746af6f2e24e6bef640e411d9bd7/Sources/PackageModel/PackageIdentity.swift#L16-L19

ikesyo:
I fixed the implementation using [CanonicalPackageLocation](https://github.com/swiftlang/swift-package-manager/blob/bc689418acad746af6f2e24e6bef640e411d9bd7/Sources/PackageModel/PackageIdentity.swift#L345-L424) instead of PackageIdentity and added the test case.

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -96,6 +96,7 @@ extension PinsStore.PinState: @retroactive Hashable {
 }
 
 public struct SwiftPMCacheKey: CacheKey {
+    public var canonicalPackageLocation: String
```
giginet, line 99: I'm worried about the remote URL being embedded. If it's used in a mirrored repository, all caches will be expired. I think it's better to be it `Optional` and it should have only for local packages.

ikesyo: Thanks for the suggestion, I addressed this in 507be71.

giginet (approved):

freddi-kit (approved):

PR #173, opened 2025-01-16
Make release workflow enable to trigger manually

ikesyo:
I'm not sure why but the release workflow for [0.27.0](https://github.com/giginet/Scipio/releases/tag/0.27.0) didn't ran. I added workflow_dispatch trigger for the workflow and would like to trigger that manually.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,12 +1,8 @@
 on: 
   release:
     types: [published]
+  # Use with only Tags
   workflow_dispatch:
-    inputs:
-      tag:
-        description: 'Tag to build and upload artifact bundle for'
-        required: true
-        default: ${{ github.ref_name }}
 
 name: Upload Artifact Bundle to Release
 env: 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,6 +1,12 @@
 on: 
   release:
     types: [created]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag to build and upload artifact bundle for'
+        required: true
+        default: ${{ github.ref_name }}
```
giginet, line 9: If this value is revision, can we update Zip files to the actual release?

.github/workflows/release.yml
```diff
@@ -1,6 +1,12 @@
 on: 
   release:
     types: [created]
```
giginet, line 3: How about adding `updated`?

giginet (approved):

.github/workflows/release.yml
ikesyo: `release` trigger: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release

- published
- unpublished
- created
- edited
- deleted
- prereleased
- released

`edited` is triggered only when the description is updated so that's not appropriate I think.

Looks like [`published`](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release) is better than `created`.

.github/workflows/release.yml
ikesyo: I think we can't upload in that case (for commit hashes).

.github/workflows/release.yml
ikesyo: > I'm not sure why but the release workflow for [0.27.0](https://github.com/giginet/Scipio/releases/tag/0.27.0) didn't ran

That was because I created the release as draft initially.

https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=created#release

> A draft was saved, or a release or pre-release was published without previously being saved as a draft.

giginet (approved):

PR #183, opened 2025-02-09
Parse `dump-pif` result and reduce XCBuildSupport dependency

giginet:
Use `swift package dump-pif` to construct PIF instead of the internal implementation of SwiftPM.

Using `dump-pif` command, we can reduce dependency and we can use stable specs.
Our final goal is to remove the `XCBuildSupport` dependency.

This PR introduces SwiftyJSON to manipulate PIFs. because it's tough to define a complete PIF using `Codable`. We only need a subset of PIF.

Instead, I  concluded the JSON manipulation into the `PIFKit` target. `ScipioKit` can manipulate PIF in type-safe.


Sources/ScipioKit/Producer/PIF/PIFGenerator.swift
```diff
@@ -1,308 +1,208 @@
...
-
-            settings[.MODULEMAP_FILE_CONTENTS] = """
-                module \(swiftTarget.c99name) {
-                    header "\(bridgingHeaderFullPath.pathString)"
-                    export *
-                }
-                """
+        if case .string(let moduleMapFileContents) = configuration.buildSettings["MODULEMAP_FILE_CONTENTS"],
+           moduleMapFileContents.contains("\(target.name)-Swift.h") {
+            resolveModuleMapPath(of: target, configuration: &configuration, sdk: sdk)
```
ikesyo, line 140:
```suggestion
        if case .string(let moduleMapFileContents) = configuration.buildSettings["MODULEMAP_FILE_CONTENTS"],
           moduleMapFileContents.contains("\(target.name)-Swift.h")
        {
            resolveModuleMapPath(of: target, configuration: &configuration, sdk: sdk)
```

would be more readable (for me).

Sources/ScipioKit/Producer/PIF/PIFGenerator.swift
```diff
@@ -1,308 +1,208 @@
...
+        configuration.buildSettings["PRODUCT_MODULE_NAME"] = "$(EXECUTABLE_NAME:c99extidentifier)"
+        configuration.buildSettings["EXECUTABLE_NAME"] = .string(target.c99Name)
+        configuration.buildSettings["TARGET_NAME"] = .string(name)
+        configuration.buildSettings["PRODUCT_BUNDLE_IDENTIFIER"] = .string(name.spm_mangledToBundleIdentifier())
+        configuration.buildSettings["CLANG_ENABLE_MODULES"] = true
+        configuration.buildSettings["DEFINES_MODULE"] = true
+        configuration.buildSettings["SKIP_INSTALL"] = false
+        configuration.buildSettings["INSTALL_PATH"] = "/usr/local/lib"
+        configuration.buildSettings["ONLY_ACTIVE_ARCH"] = false
+        configuration.buildSettings["SDKROOT"] = .string(sdk.settingValue)
```
ikesyo, line 98: `SDKROOT` is added, for what reason is that required?

Sources/ScipioKit/Producer/PIF/PIFGenerator.swift
```diff
@@ -1,308 +1,208 @@
...
     private let buildOptions: BuildOptions
     private let buildOptionsMatrix: [String: BuildOptions]
+    private let executor: any Executor
     private let fileSystem: any FileSystem
 
     init(
-        package: DescriptionPackage,
+        packageName: String,
+        packageLocator: some PackageLocator,
         buildParameters: BuildParameters,
```
ikesyo, line 21: Because we don't use `PIFBuilder.generatePIF` anymore, only `buildParameters.toolchain.toolchainLibDir` is needed. How about passing the specific value instead of the entire `BuildParameters`?

Ryu0118 (approved):

Sources/PIFKit/PIFManipulator.swift
```diff
@@ -0,0 +1,51 @@
+import Foundation
+import SwiftyJSON
+
+/// Manipulates PIF JSON data
+package class PIFManipulator {
```
giginet, line 5:
```suggestion
package final class PIFManipulator {
```

Sources/ScipioKit/Producer/PIF/PIFGenerator.swift
giginet: dumppif for `.xcodeproj` generates this value, so I added, but we needn't set the value. I'll revert this.

Sources/PIFKit/BuildConfiguration.swift
```diff
@@ -0,0 +1,101 @@
...
+        package var buildSettings: [String: MacroExpressionValue]
+
+        package init(buildSettings: [String: MacroExpressionValue]) {
+            self.buildSettings = buildSettings
+        }
+    }
+
+    package var name: String
+    package var buildSettings: [String: MacroExpressionValue]
+    package var baseConfigurationFileReferenceGUID: String?
```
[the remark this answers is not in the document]
ikesyo: Removed in this commit: https://github.com/giginet/Scipio/pull/183/commits/b6aa079fc95f4b1813d0f14118d9f25eb08f7588

ikesyo (approved):

PR #185, opened 2025-02-15
Add feature to strip DWARF symbols from the final binary

giginet:
This PR introduces the new flag `--strip-dwarf-symbols`.

To share the clang static built frameworks from the other machine, LLDB attempts to load *.pcm, but it's not shared among the build machines.
This option will strip DWAFR information from the final built binary to avoid this situation.

It's better to use `DEBUG_INFORMATION_FORMAT = dwarf-with-dsym` instead.


Sources/ScipioKit/Producer/PIF/PIFCompiler.swift
```diff
@@ -78,11 +80,17 @@ struct PIFCompiler: Compiler {
...
             do {
-                try await xcBuildClient.buildFramework(
+                let frameworkBundlePath = try await xcBuildClient.buildFramework(
                     sdk: sdk,
                     pifPath: pifPath,
                     buildParametersPath: buildParametersPath
                 )
+
+                if buildOptions.stripDWARFSymbols {
+                    logger.info("🐛Stripping debug symbols")
```
ikesyo, line 90:
```suggestion
                    logger.info("🐛 Stripping debug symbols")
```

Sources/ScipioKit/Producer/PIF/PIFCompiler.swift
```diff
@@ -78,11 +80,17 @@ struct PIFCompiler: Compiler {
...
+                    let binaryPath = frameworkBundlePath.appending(component: buildProduct.target.c99name)
```
ikesyo, line 91: > `buildProduct.target.c99name`

This is the same as https://github.com/giginet/Scipio/blob/e678d7e3b9b6e354457716c4a184138681c36ba4/Sources/ScipioKit/Producer/PIF/FrameworkComponentsCollector.swift#L74-L82 so it's okay 🙆

Sources/scipio/Options.swift
```diff
@@ -45,6 +45,11 @@ struct BuildOptionGroup: ParsableArguments {
     @Flag(name: [.customShort("f", allowingJoined: false), .long],
           help: "Whether overwrite generated frameworks or not")
     var overwrite: Bool = false
+
+    @Flag(name: [.customLong("--strip-dwarf-symbols")],
+          inversion: .prefixedNo,
+          help: "Whether to strip DWARF symbol from built binary or not")
+    var shouldStripDWARFSymbols: Bool = false
```
ikesyo, line 52: Let's put this between `shouldEnableLibraryEvolution` and `shouldOnlyUseVersionsFromResolvedFile`.

Sources/scipio/scipio.docc/prepare-cache-for-applications.md
```diff
@@ -96,6 +96,7 @@ All XCFrameworks are generated into `MyAppDependencies/XCFramework` by default.
 | -\-cache-policy                         | How to reuse built frameworks                                       | project                    |
 | -\-enable-library-evolution             | Whether to enable Library Evolution feature or not                  | -                          |
 | -\-only-use-versions-from-resolved-file | Whether to disable updating Package.resolved automatically          | false                      |
+| -\-strip-dwarf-symbols                  | Whether to strip DWARF symbols from built binary or not              | false                      |
```
ikesyo, line 99:
```suggestion
| -\-enable-library-evolution             | Whether to enable Library Evolution feature or not                  | -                          |
| -\-strip-dwarf-symbols                  | Whether to strip DWARF symbols from built binary or not              | false                      |
| -\-only-use-versions-from-resolved-file | Whether to disable updating Package.resolved automatically          | false                      |
```

Tests/ScipioKitTests/DWARFSymbolStripperTests.swift
```diff
@@ -0,0 +1,49 @@
...
+
+        let binaryPath = frameworkOutputDir.appending(
+            components: "some_lib.xcframework", "ios-arm64", "some_lib.framework", "some_lib"
+        )
+        #expect(fileManager.fileExists(atPath: binaryPath.path(percentEncoded: false)))
+
+        let executor = ProcessExecutor()
+        let dwarfDumpExecutionResult = try await executor.execute(
+            "/usr/bin/xcrun", "dwarfdump", "--debug-info", binaryPath.path(percentEncoded: false)
+        )
```
ikesyo, line 41:
```suggestion
        let binaryPath = frameworkOutputDir.appending(
            components: "some_lib.xcframework", "ios-arm64", "some_lib.framework", "some_lib"
        ).path(percentEncoded: false)
        #expect(fileManager.fileExists(atPath: binaryPath))

        let executor = ProcessExecutor()
        let dwarfDumpExecutionResult = try await executor.execute(
            "/usr/bin/xcrun", "dwarfdump", "--debug-info", binaryPath
        )
```

Tests/ScipioKitTests/DebugSymbolTests.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 @testable import ScipioKit
 import Testing
 
+@Suite(.serialized)
 struct DwarfExtractorTests {
```
ikesyo, line 6: Why these `.serialized` traits are added / required?

giginet: swift-testing runs tests in parallel in default. As I investigate, it may cause CI hung. So it's safe to annotate `serialized` for all test cases to execute the build system. It may not be thread-safe.

ikesyo (approved):

Sources/scipio/Options.swift
```diff
@@ -38,6 +38,11 @@ struct BuildOptionGroup: ParsableArguments {
           help: "Whether to enable Library Evolution feature or not")
     var shouldEnableLibraryEvolution = false
 
+    @Flag(name: [.customLong("--strip-dwarf-symbols")],
+          inversion: .prefixedNo,
+          help: "Whether to strip DWARF symbol from built binary or not")
+    var shouldStripDWARFSymbols: Bool = false
```
ikesyo, line 44: We should describe that the option is only for static frameworks (both on its name and help).

Sources/scipio/Options.swift
```diff
@@ -38,6 +38,11 @@ struct BuildOptionGroup: ParsableArguments {
           help: "Whether to enable Library Evolution feature or not")
     var shouldEnableLibraryEvolution = false
 
+    @Flag(name: [.customLong("--strip-dwarf-symbols")],
```
ikesyo, line 41:
```suggestion
    @Flag(name: [.customLong("--strip-static-lib-dwarf-symbols")],
```

Is this redundant?

ikesyo (approved):
LGTM 👏

ikesyo:
Unfortunately https://github.com/giginet/Scipio/commit/43c29d3045579806f7efc177b83481f967eca014 was deleted by the last rebase. 😢

PR #191, opened 2025-02-25
Surpress logging of stripping debug information

giginet:
Make log level to debug


Sources/ScipioKit/Producer/PIF/PIFCompiler.swift
```diff
@@ -87,7 +87,7 @@ struct PIFCompiler: Compiler {
                 )
 
                 if buildOptions.stripDWARFSymbols && buildOptions.frameworkType == .static {
-                    logger.info("🐛 Stripping debug symbols")
+                    logger.debug("🐛 Stripping debug symbols of \(target.name)\(sdk.displayName)")
```
ikesyo, line 90:
```suggestion
                    logger.debug("🐛 Stripping debug symbols of \(target.name) (\(sdk.displayName))")
```

ikesyo (approved):

PR #192, opened 2025-03-03
Support swift-package-manager `release/6.1` branch

ikesyo:
The branch can be built by Swift 6.0 compiler and works as expected (All tests are passed).

### DIFF
--- a/Sources/ScipioKit/Producer/PIF/PIFGenerator.swift
+++ b/Sources/ScipioKit/Producer/PIF/PIFGenerator.swift
@@ -207,9 +207,8 @@ private struct PIFLibraryTargetModifier {
 
         let toolchainLibDir = (try? buildParameters.toolchain.toolchainLibDir) ?? .root
 
-        settings[.PRODUCT_NAME] = "$(EXECUTABLE_NAME:c99extidentifier)"
-        settings[.PRODUCT_MODULE_NAME] = "$(EXECUTABLE_NAME:c99extidentifier)"
-        settings[.EXECUTABLE_NAME] = c99Name
+        settings[.PRODUCT_NAME] = c99Name
+        settings[.PRODUCT_MODULE_NAME] = c99Name
         settings[.TARGET_NAME] = name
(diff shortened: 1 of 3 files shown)


Sources/ScipioKit/Producer/PIF/PIFGenerator.swift
```diff
@@ -207,9 +207,8 @@ private struct PIFLibraryTargetModifier {
 
         let toolchainLibDir = (try? buildParameters.toolchain.toolchainLibDir) ?? .root
 
-        settings[.PRODUCT_NAME] = "$(EXECUTABLE_NAME:c99extidentifier)"
-        settings[.PRODUCT_MODULE_NAME] = "$(EXECUTABLE_NAME:c99extidentifier)"
-        settings[.EXECUTABLE_NAME] = c99Name
```
giginet, line 212: Is it OK to remove this?

If the migration of #183 is merged, we can use any parameters.

ikesyo: https://developer.apple.com/documentation/Xcode/build-settings-reference#EXECUTABLENAME

There may be an effect to the generated Info.plist's `CFBundleExecutable` key, but that is tested here: https://github.com/giginet/Scipio/blob/ba192cb69f899bff568aafd50a7520f79d265434/Tests/ScipioKitTests/RunnerTests.swift#L100

So I think it would be okay.

ikesyo: It seems that the `EXECUTABLE_NAME` is by default `$(EXECUTABLE_PREFIX)$(PRODUCT_NAME)$(EXECUTABLE_SUFFIX)`. In SwiftPM 6.1, `EXECUTABLE_PREFIX` is used only for `func addLibraryTarget(for product: ResolvedProduct)` which we don't depend on (we depend on `ResolvedModule` instead). So we will not be affected by SwiftPM changes.

giginet (approved):

PR #200, opened 2025-03-22
Introduce dependency graph

mtj0928:
Scipio manages module dependencies as an `OrderedSet` sorted by topological order. 
While this works, it makes it difficult to optimize the build process — for example, adding macro support or enabling parallel builds — because the data structure loses information about which modules depend on which other modules.

To address this issue, I’ve introduced a new data structure: a dependency graph. 
This data structure retains information about module dependencies, making it possible to improve the build process.

I haven’t added any new features to Scipio in this PR, but these changes will help support future development.


mtj0928:
Hmm..., when I run the tests with Xcode 16.3 in my local environment, all tests were passed.
However, I switched the version to 16.2, and the tests were failed....

And the issues occurs in the main branch to...

ikesyo:
> Hmm..., when I run the tests with Xcode 16.3 in my local environment, all tests were passed.
> However, I switched the version to 16.2, and the tests were failed....

That's because `ResolvedModule.ID` contains `BuildTriple` on [Swift 6.0](https://github.com/swiftlang/swift-package-manager/blob/bc689418acad746af6f2e24e6bef640e411d9bd7/Sources/PackageGraph/Resolution/ResolvedModule.swift#L296-L310) but doesn't on [Swift 6.1](https://github.com/swiftlang/swift-package-manager/blob/1fc90e29029bfeafe3550ccf08f74a86a11baa23/Sources/PackageGraph/Resolution/ResolvedModule.swift#L277-L290).

This is related to these:
- https://github.com/giginet/Scipio/blob/b6426bd3b0c42b6a1c3722ab92357d6cb82d4314/Package.swift#L7-L12
- https://github.com/giginet/Scipio/blob/b6426bd3b0c42b6a1c3722ab92357d6cb82d4314/Sources/ScipioKit/DescriptionPackage.swift#L138-L154

We need to avoid `ResolvedModule.ID` on Swift 6.0 but it may be difficult 🤔 

- https://github.com/giginet/Scipio/blob/083fc33e8ac17a86c3914592e39fa413959e6f98/Sources/ScipioKit/DescriptionPackage.swift#L168-L172
- https://github.com/giginet/Scipio/blob/083fc33e8ac17a86c3914592e39fa413959e6f98/Sources/ScipioKit/DescriptionPackage.swift#L233-L241

ikesyo:
We can use this API for retrieving the package of a resolved module: https://github.com/swiftlang/swift-package-manager/blob/db9fef21d000dd475816951d52f8d32077939e81/Sources/PackageGraph/ModulesGraph.swift#L152-L155

`BuildProductsResolver` has `descriptionPackage` so we can use it in `resolveBuildProductDependencyGraph()`.

mtj0928:
@ikesyo Thank you for sharing the cause.
I used module name for Swift 6.0 instead of module id.
It's not accurate, but the original implementation also used module name for resolve modules.
So, the behaviors were not changed.

However, there is still a failed test case, `testWithResourcePackage`. 
The test case also failed in main branch...

Sources/ScipioKit/DependencyGraph.swift
```diff
@@ -0,0 +1,188 @@
...
+            childNodes.forEach { childNode in node.appendToChildren(childNode) }
+        }
+        let rootNodes = nodes.filter { node in node.parents.isEmpty }
+        return DependencyGraph(rootNodes: rootNodes, allNodes: nodes)
+    }
+}
+
+extension DependencyGraph {
+    /// All leaf nodes in this graph.
+    func leafs() -> [Node] {
```
giginet, line 44: I prefer it to be a computed property.

Sources/ScipioKit/DependencyGraph.swift
```diff
@@ -0,0 +1,188 @@
+import Foundation
```
giginet, line 1: Could you move this and `WeakReference` implementations to `ScipioKit/DependencyGraph/DependencyGraph.swift`

giginet (approved):

Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -112,9 +112,9 @@ struct DescriptionPackage: PackageLocator {
 }
 
 extension DescriptionPackage {
-    func resolveBuildProducts() throws -> OrderedSet<BuildProduct> {
```
ikesyo, line 115: We should be able to remove `import Collections` in this file.

ikesyo (approved):
LGTM 👏

Sources/ScipioKit/DescriptionPackage.swift
mtj0928: I removed the import, thank you!

ikesyo (approved):

giginet (approved):

PR #184, opened 2025-02-13
Leave cache of frameworks that are successfully built

S-Shimotori:
Closes #153 

This PR will fix the issue reported by #153.

# Changes

## FrameworkProducer

`processAllTargets(buildProducts:)` will cache XCFrameworks that are successfully built.
Scipio users won't have to rebuild all frameworks even if Scipio fails to build some frameworks.

## PartialCacheTests

`PartialCacheTests` checks that `processAllTargets(buildProducts:)` works well.

Tests/ScipioKitTests/Resources/Fixtures/PartialCacheTestPackage provides two libraries, `Base` and `Bad.`
`Bad` depends on `Base` and cannot be compiled.

PR #203, opened 2025-04-01
Fix issue causing valid local cache to be discarded and external cache to be fetched

Ryu0118:
In this PR, I fixed the issue causing valid local cache to be discarded and remote cache to be fetched.

The problem was that version files were created only for built targets, not for all targets. Previously, for caches obtained from local or remote sources, we did not generate version files. As a result, Scipio mistakenly marked valid local cache as invalid and fetched the remote cache unnecessarily.

### DIFF
--- a/Sources/ScipioKit/Producer/FrameworkProducer.swift
+++ b/Sources/ScipioKit/Producer/FrameworkProducer.swift
@@ -85,6 +85,8 @@ struct FrameworkProducer {
             )
         }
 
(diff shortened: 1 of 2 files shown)


giginet (approved):

ikesyo:
ref: https://github.com/giginet/Scipio/pull/184/files#diff-eb9630eddbe115fb1a8e1dc1e1499ae8f9e3b7739309721b57f433525fb8f2efR137-R139

ikesyo (approved):

PR #208, opened 2025-05-16
Migrate Package Resolver from SwiftPM

Ryu0118:
I implemented a custom logic to resolve packages, which was previously dependent on SwiftPM.
This new logic is built using the outputs of `dump-package` and `show-dependencies`.

While this PR does not completely remove the SwiftPM dependency, it is a step in that direction.
To fully decouple from SwiftPM, we still need to remove its usage from `UserToolchain`, `CacheSystem`, `FileSystem`, `AbsolutePath`, etc.


Sources/ScipioKit/PackageKind+enabledKinds.swift
```diff
@@ -0,0 +1,7 @@
+import PackageManifestKit
+
+extension Target.TargetKind {
+    static var enabledKinds: [Self] {
```
giginet, line 4: It's difficult to understand what `enabled` means.

It's better to be function of `FrameworkProducer` rather than the extension.

`func shouldBuild(_ target: Target.TargetKind)`

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -4,9 +4,10 @@ import Basics
 import struct TSCUtility.Version
 import Algorithms
 // We may drop this annotation in SwiftPM's future release
-@preconcurrency import PackageGraph
-import PackageModel
+@preconcurrency import class PackageGraph.PinsStore
```
giginet, line 7: We have to replace PinsStore in the future.

Sources/ScipioKit/Producer/PIF/FrameworkComponentsCollector.swift
```diff
@@ -146,7 +145,14 @@ struct FrameworkComponentsCollector {
     }
 
     private func generatedResourceBundlePath() -> TSCAbsolutePath? {
-        guard let bundleName = buildProduct.target.underlying.bundleName else { return nil }
+        let bundleName: String? =
+            if buildProduct.target.underlying.resources.isEmpty {
+                nil
+            } else {
+                buildProduct.package.manifest.name + "_" + buildProduct.target.underlying.name
+            }
```
giginet, line 153: Could you move this logic to extension of underlying target?

Sources/ScipioKit/Producer/PIF/FrameworkComponentsCollector.swift
```diff
@@ -198,24 +204,29 @@ struct FrameworkComponentsCollector {
...
-        let notSymlinks = publicHeaders.filter { !fileSystem.isSymlink($0) }
-            .map { $0.scipioAbsolutePath }
-        let symlinks = publicHeaders.filter { fileSystem.isSymlink($0) }
+        let headerExtensions = ["h", "hh", "hpp", "h++", "hp", "hxx", "H", "ipp", "def"]
+
+        let publicHeaders: [URL] = FileManager.default
+            .enumerator(at: includeDir, includingPropertiesForKeys: nil)?
+            .compactMap { $0 as? URL }
+            .filter { headerExtensions.contains($0.pathExtension) }
+            ?? []
```
giginet, line 217: Could you move this logic to target model?

Sources/ScipioKit/Resolver/ManifestLoader.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+import PackageManifestKit
+
+/// Loads the `Manifest` of a Swift package
+struct ManifestLoader: @unchecked Sendable {
```
giginet, line 5: Why should be `@unechecked Sendable`?

Sources/ScipioKit/Resolver/PackageResolver.swift
```diff
@@ -0,0 +1,532 @@
+import Foundation
+import PackageManifestKit
+import OrderedCollections
+import AsyncOperations
+import Basics
+
+actor PackageResolver {
```
giginet, line 7: Read it later

Sources/ScipioKit/Resolver/ResolvedModels.swift
```diff
@@ -0,0 +1,233 @@
...
+            description = url.absoluteString
+        case .remoteSourceControl(let string):
+            description = string
+        case .registry(let string):
+            description = string
+        }
+    }
+}
+
+struct ResolvedPackage: Identifiable {
```
giginet, line 40: Should be Sendable?

Sources/ScipioKit/Resolver/ResolvedModels.swift
```diff
@@ -0,0 +1,233 @@
...
+        self.path = path
+        self.targets = targets
+        self.products = products
+    }
+}
+
+extension ResolvedPackage {
+    var canonicalPackageLocation: CanonicalPackageLocation {
+        CanonicalPackageLocation(
+            URL(fileURLWithPath: path)
```
giginet, line 75: `URL(filePath:)`

Sources/ScipioKit/Resolver/ResolvedModels.swift
Ryu0118: Since ResolvedPackage is internal, isn't it implicitly Sendable?

Sources/ScipioKit/PackageKind+enabledKinds.swift
Ryu0118: `enabledKinds` is also used in `PackageResolver`, so I prefer not to include it within `FrameworkProducer`.

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -233,14 +231,14 @@ struct CacheSystem: Sendable {
...
+        let localPackageCanonicalLocation: String? = switch package.resolvedPackageKind {
         case .fileSystem, .localSourceControl:
-            package.manifest.canonicalPackageLocation.description
+            package.canonicalPackageLocation.description
         case .root, .remoteSourceControl, .registry:
             nil
         }
 
-        let pin = try retrievePin(package: package)
+        let pin = try retrievePinState(package: package)
```
ikesyo, line 241: This should be named as `pinState`.

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -294,7 +293,8 @@ public struct VersionFileDecoder {
 }
 
 extension ResolvedPackage {
-    fileprivate func makePinFromRevision() -> PinsStore.Pin? {
+    fileprivate func makePinFromRevision() throws -> PinsStore.PinState? {
```
ikesyo, line 296:
```suggestion
    fileprivate func makePinStateFromRevision() throws -> PinsStore.PinState? {
```

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -303,15 +303,38 @@ extension ResolvedPackage {
...
-            )
+        return .version(
+            version,
+            revision: try? repository.getCurrentRevision().identifier
         )
     }
 }
+
+fileprivate extension PackageManifestKit.PackageKind {
+    var spmPackageKind: PackageReference.Kind {
```
ikesyo, line 314: Looks like the extension is not used in this file so could be removed, right?

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -261,11 +259,12 @@ struct CacheSystem: Sendable {
         )
     }
 
-    private func retrievePin(package: ResolvedPackage) throws -> PinsStore.Pin {
-        guard let pin = pinsStore.pins[package.identity] ?? package.makePinFromRevision() else {
-            throw Error.revisionNotDetected(package.manifest.displayName)
+    private func retrievePinState(package: ResolvedPackage) throws -> PinsStore.PinState {
+        guard let pinState = package.pinState?.spmPinState ?? (try? package.makePinFromRevision()) else {
```
ikesyo, line 263: Can't we avoid using `try?`?

Sources/ScipioKit/Producer/BinaryExtractor.swift
```diff
@@ -1,26 +1,29 @@
...
+    var descriptionPackage: DescriptionPackage
     var outputDirectory: URL
     var fileSystem: any FileSystem
 
     @discardableResult
-    func extract(of binaryTarget: ScipioBinaryModule, overwrite: Bool) throws -> URL {
-        let sourcePath = binaryTarget.artifactPath
+    func extract(of binaryTarget: ResolvedModule, overwrite: Bool) throws -> URL {
+        guard case let .binary(binaryLocation) = binaryTarget.resolvedModuleType else {
+            preconditionFailure("\(#function) must be called with a binary target")
```
ikesyo, line 12: The log message should contain the target's name and its actual module type to make the debugging easier.

Sources/ScipioKit/SwiftPM/CannonicalPackageLocation.swift
ikesyo, line 1: It would be better to add README in `Sources/ScipioKit/SwiftPM` directory to describe that the directory contains the copies (or slightly modified versions) of some source files from swift-package-manager.

Tests/ScipioKitTests/DescriptionPackageTests.swift
```diff
@@ -76,17 +76,13 @@ final class DescriptionPackageTests: XCTestCase {
         let graph = try package.resolveBuildProductDependencyGraph()
             .map { $0.target.name }
 
-        let myPluginNode = try XCTUnwrap(graph.rootNodes.first)
-        XCTAssertEqual(myPluginNode.value, "MyPlugin")
-
-        let executableTargetNode = try XCTUnwrap(myPluginNode.children.first)
-        XCTAssertEqual(executableTargetNode.value, "ExecutableTarget")
```
ikesyo, line 83: Is is actually oaky to remove these assertions?

ref:
- #200

Sources/ScipioKit/Resolver/ResolvedModels.swift
```diff
@@ -0,0 +1,233 @@
...
+        switch packageKind {
+        case .root(let url), .fileSystem(let url), .localSourceControl(let url):
+            description = url.absoluteString
+        case .remoteSourceControl(let string):
+            description = string
+        case .registry(let string):
+            description = string
+        }
+    }
+}
```
ikesyo, line 38: I think that the implementation will not have correct dependency mirroring support which should exist in the original SwiftPM implementation: https://github.com/swiftlang/swift-package-manager/blob/1cbb77aa98ab419c57d0d78287ac2a3a4202c1d7/Sources/PackageModel/IdentityResolver.swift#L25-L77

Removing dependency mirroring support will be a breaking change or may make the tool unable to use in some real-world use cases. Is it an intentional decision? Is it okay to remove that support?

@Ryu0118 @giginet

Tests/ScipioKitTests/DescriptionPackageTests.swift
Ryu0118: Yes, since `PackageResolver` filters targets to only `regular` and `binary` types, I removed this test assertion.


https://github.com/Ryu0118/Scipio/blob/5f4d76b5e2a011f8a04948c06624fd6ae28af701/Sources/ScipioKit/Resolver/PackageResolver.swift#L74
https://github.com/Ryu0118/Scipio/blob/5f4d76b5e2a011f8a04948c06624fd6ae28af701/Sources/ScipioKit/Resolver/PackageResolver.swift#L206

Sources/ScipioKit/Resolver/ResolvedModels.swift
giginet: It's better to note the concern for now. However, we needn't resolve this issue in this PR.

First, we should investigate whether `dump-package` contains mirroring information.

NOTICE.txt
ikesyo, line 1: Is this file needed in addition to `Sources/ScipioKit/SwiftPM/LICENSE`? This repository itself is not the SwiftPM project so placing this file to the repository root is confusing. 🤔

Sources/ScipioKit/Resolver/PackageResolver.swift
```diff
@@ -0,0 +1,544 @@
...
+        if let cachedModuleType = cachedModuleType[target] {
+            return cachedModuleType
+        }
+
+        let resolvedModuleType = ModuleTypeResolver(
+            fileSystem: fileSystem,
+            rootPackageDirectory: packageDirectory,
+            target: target,
+            dependencyPackage: dependencyPackage
+        ).resolve()
```
ikesyo, line 288: Looks like we can hold an instance of ModuleTypeResolver as a property of PackageResolver and pass `target` and `dependencyPackage` arguments to `resolve` method.

```suggestion
        let resolvedModuleType = moduleTypeResolver.resolve(
            target: target,
            dependencyPackage: dependencyPackage
        )
```

Sources/ScipioKit/Resolver/PackageResolver.swift
```diff
@@ -0,0 +1,544 @@
...
+        }
+        self.resolvedPackageKinds.merge(actualPackageKinds) { _, new in new }
+    }
+
+    private func shouldBuild(_ target: Target.TargetKind) -> Bool {
+        target == .regular || target == .binary
+    }
+}
+
+private struct ModuleTypeResolver {
```
ikesyo, line 337: This file is too-large so splitting into smaller chunks by extracting each component into its own files will be better (it would be okay to do that refactoring in a separate PR). Then it will be easier to test each component.

- ModuleTypeResolver
- PackageResolveExecutor
- ShowDependenciesParser

The types will be nested in PackageResolver.

```swift
extension PackageResolver {
    struct ModuleTypeResolver {
        ...
    }
}
```

ikesyo (approved):

NOTICE.txt
ikesyo: Thanks for the clarification!

PR #212, opened 2025-06-03
Migrate PinsStore.PinState from SwiftPM

Ryu0118:
I migrated `PinsStore.PinState` from SwiftPM.
To keep cache compatibility, I didn’t use `Pin.State` in `Sources/Resolver/ResolvedModels.swift`. Instead, I used `PinsStore.PinState` from SwiftPM and redefined it as `SPMPinState` in `Sources/SwiftPM/PinState.swift`.


Sources/ScipioKit/SwiftPM/PinState.swift
```diff
@@ -0,0 +1,85 @@
...
+// See http://swift.org/LICENSE.txt for license information
+// See http://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+// ===----------------------------------------------------------------------===//
+//
+// NOTE: This file has been modified from the original version.
+//
+// ===----------------------------------------------------------------------===//
+
+public enum SPMPinState: Equatable, Sendable, CustomStringConvertible {
```
giginet, line 17: I feel it's enough just to be `PinState`

Sources/ScipioKit/SwiftPM/Version.swift
```diff
@@ -0,0 +1,300 @@
+// swiftlint:disable:next blanket_disable_command
+// swiftlint:disable all
```
ikesyo, line 2: We should add `excluded` entry to `.swiftlint.yml` for this directory instead of adding disable comments.

https://github.com/giginet/Scipio/blob/c96f2605ca958d7fdae80bb1a892918049d023e9/.swiftlint.yml#L4-L6

Sources/ScipioKit/SwiftPM/PinState.swift
ikesyo: > To keep cache compatibility, I didn’t use `Pin.State` in `Sources/Resolver/ResolvedModels.swift`. Instead, I used `PinsStore.PinState` from SwiftPM and redefined it as `SPMPinState` in `Sources/SwiftPM/PinState.swift`.

If Scipio version is changed (updated) cache keys are invalidated anyway so we don't need to such compatibility I think 👀

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -324,3 +269,43 @@ fileprivate extension Pin.State {
...
+    func fetchCurrentRevision() async throws -> String {
+        let arguments = [
+            "/usr/bin/xcrun",
+            "git",
+            "-C",
+            path.path(percentEncoded: false),
+            "rev-parse",
+            "--verify",
+            "HEAD",
+        ]
```
ikesyo, line 306: ref: https://github.com/swiftlang/swift-package-manager/blob/e23a72d0f61e1cbc79adf030803881bf2da24bce/Sources/SourceControl/GitRepository.swift#L612-L621

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -324,3 +269,43 @@ fileprivate extension Pin.State {
...
+    func fetchCurrentTag() async throws -> String {
+        let arguments = [
+            "/usr/bin/xcrun",
+            "git",
+            "-C",
+            path.path(percentEncoded: false),
+            "describe",
+            "--exact-match",
+            "--tags",
+        ]
```
ikesyo, line 291: ref: https://github.com/swiftlang/swift-package-manager/blob/e23a72d0f61e1cbc79adf030803881bf2da24bce/Sources/SourceControl/GitRepository.swift#L623-L632

Sources/ScipioKit/SwiftPM/PinState.swift
giginet: Yes, as ikesyo-san said, we needn't consider backward compatibility. All caches will expire

.swiftlint.yml
```diff
@@ -4,6 +4,7 @@ included:
 excluded:
   - .build
   - Tests/ScipioKitTests/Resources/Fixtures
+  - Sources/ScipioKit/SwiftPM/Version.swift
```
giginet, line 7: How about for other SwiftPM sources?

giginet (approved):

.swiftlint.yml
ikesyo:
```suggestion
  - Sources/ScipioKit/SwiftPM
```

giginet (approved):

PR #214, opened 2025-06-10
Improve `ModuleTypeResolver.resolveTargetFullPath` to support case-insensitive matching of default module directories

Ryu0118:
Previously, `ModuleTypeResolver` could only resolve modules located under the following directory structures:

* `Sources/<target name>`
* `src/<target name>`

However, SwiftPM treats these directory names in a case-insensitive manner. This means that modules placed under directories like `sources/` or `SRC/`, which are valid in SwiftPM, were not recognized by `ModuleTypeResolver`.

Additionally, in packages that contain only a single module, SwiftPM allows placing source files directly under `Sources/` (e.g. `Sources/main.swift`) without a subdirectory. The previous implementation did not account for this layout either.

This PR updates `resolveTargetFullPath` to handle:

* Case-insensitive matching of standard source directories (`sources`, `source`, `src`, `srcs`)
* Single-module packages where source files are placed directly under the root source directory

It also adds tests to cover various valid module directory layouts.

### DIFF
--- a/Sources/ScipioKit/DescriptionPackage.swift
+++ b/Sources/ScipioKit/DescriptionPackage.swift
@@ -33,26 +33,6 @@ struct DescriptionPackage: PackageLocator {
         Set(manifest.platforms?.map(\.platformName).compactMap(SDK.init(platformName:)) ?? [])
     }
 
-    private static func makeManifest(
-        packageDirectory: TSCAbsolutePath,
-        jsonDecoder: JSONDecoder,
-        executor: some Executor
-    ) async throws -> PackageManifestKit.Manifest {
-        let commands = [
-            "/usr/bin/xcrun",
-            "swift",
-            "package",
-            "dump-package",
-            "--package-path",
-            packageDirectory.pathString,
-        ]
-
-        let manifestString = try await executor.execute(commands).unwrapOutput()
-        let manifest = try jsonDecoder.decode(PackageManifestKit.Manifest.self, from: manifestString)
-
-        return manifest
-    }
-
     /// Make DescriptionPackage from a passed package directory
     /// - Parameter packageDirectory: A path for the Swift package to build
     /// - Parameter mode: A Scipio running mode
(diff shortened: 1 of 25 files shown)


Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -33,26 +33,6 @@ struct DescriptionPackage: PackageLocator {
         Set(manifest.platforms?.map(\.platformName).compactMap(SDK.init(platformName:)) ?? [])
     }
 
-    private static func makeManifest(
```
giginet, line 36: Is this no longer unused?

Tests/ScipioKitTests/VariousModulePathsTests.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+import Testing
+import TSCBasic
+import PackageManifestKit
+@testable import ScipioKit
+
+struct VariousModulePathsTests {
+    private static let fixturePath = URL(fileURLWithPath: #filePath)
```
giginet, line 8:
```suggestion
    private static let fixturePath = URL(filePath: #filePath)
```

Tests/ScipioKitTests/VariousModulePathsTests.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    private static let packages: [URL] = [
+        fixturePath.appending(components: "WithExplicitPath"),
+        fixturePath.appending(components: "WithoutPath_LowercaseDir_srcs"),
+        fixturePath.appending(components: "WithoutPath_LowercaseDir_source"),
+        fixturePath.appending(components: "WithoutPath_SingleModule"),
+        fixturePath.appending(components: "WithoutPath_LowercaseDir_sources"),
+        fixturePath.appending(components: "WithoutPath_UppercaseMix"),
+        fixturePath.appending(components: "WithoutPath_LowercaseDir_src"),
+    ]
```
giginet, line 20:
```suggestion
    private static let packages: [URL] = [
        "WithExplicitPath",
        "WithoutPath_LowercaseDir_srcs",
        "WithoutPath_LowercaseDir_source",
        "WithoutPath_SingleModule",
        "WithoutPath_LowercaseDir_sources",
        "WithoutPath_UppercaseMix",
        "WithoutPath_LowercaseDir_src",
    ].map { fixturePath.appending(components: $0) }
```

giginet (approved):
Thanks!!!

giginet (approved):

PR #222, opened 2025-06-23
Fix issue where xcassets and CoreData model are not included in Resource Bundle

Ryu0118:
In the current version of Scipio, there is a problem where `xcassets` and CoreData models are not included in the resource bundle when building a module that contains resources.

This issue is caused by a difference in behavior between the `swiftpm` and `xcode` build systems when executing `dump-pif`. 
we can see the relevant implementation details in the following links:

- https://github.com/swiftlang/swift-package-manager/blob/f6be6d36c0ae78a9f08309de6364623d589603de/Sources/CoreCommands/SwiftCommandState.swift#L498C64-L498C83
- https://github.com/swiftlang/swift-package-manager/blob/f6be6d36c0ae78a9f08309de6364623d589603de/Sources/CoreCommands/SwiftCommandState.swift#L1128
- https://github.com/swiftlang/swift-package-manager/blob/f6be6d36c0ae78a9f08309de6364623d589603de/Sources/PackageLoading/TargetSourcesBuilder.swift#L755

To fix this, this PR adds the `--build-system xcode` option when running `dump-pif`.
With this change, `xcassets` and CoreData models will now be correctly included in the generated resource bundle.


giginet (approved):
🚀

ikesyo:
Context: https://github.com/giginet/Scipio/pull/208#issuecomment-2979033305

ikesyo (approved):
LGTM 👍

PR #223, opened 2025-06-24
Bump `swift-nio` & `swift-nio-ssl` in integration test fixtures and update tests

Ryu0118:
This PR updates the test fixtures used in ScipioKit’s integration tests to use the latest versions of swift-nio and swift-nio-ssl.
Originally, the fixtures specified swift-nio with from: 2.0.0, but when version 2.83.0 was used locally, some tests failed due to API changes.
To fix this, the fixtures are now updated to 2.83.0, and the test code has been modified to match the updated APIs.


giginet (approved):
🌸

PR #228, opened 2025-07-07
Remove unnecessary resource copy of test fixtures

Ryu0118:
I removed an unnecessary resource copy configuration for the test target from Package.swift.

Copying the large set of test fixtures was causing significant delays before the start of test execution.
Since the tests do not fetch fixtures via the Bundle, but instead access them directly through the file system using explicit paths, the resource copy was redundant.


ikesyo (approved):
LGTM 🧹

PR #229, opened 2025-07-11
Expose ProcessExecutor via `@_spi(Internals)`

Ryu0118:
This PR makes the existing `ProcessExecutor` in ScipioKit publicly accessible under `@_spi(Internals)`.

### Motivation

Custom build pipelines in ScipioKit sometimes require executing shell commands.  
However, `TSC.Process` is not `Sendable` and cannot be abstracted via protocols.

To address these limitations, `ProcessExecutor` was previously introduced as a `Sendable` and protocol-oriented alternative.

By exposing it through `@_spi(Internals)`, this PR enables executing arbitrary commands in a `Sendable` and testable way, simplifying the development of custom pipelines in ScipioKit.

### DIFF
--- a/Sources/ScipioKit/Executor.swift
+++ b/Sources/ScipioKit/Executor.swift
@@ -9,21 +9,25 @@ private actor ProcessOutputBuffer {
     }
 }
 
-protocol Executor {
+@_spi(Internals)
+public protocol Executor {
     @discardableResult
     func execute(_ arguments: [String]) async throws -> ExecutorResult
 }
 
-protocol ErrorDecoder: Sendable {
+@_spi(Internals)
+public protocol ErrorDecoder: Sendable {
     func decode(_ result: ExecutorResult) throws -> String?
 }
 
-enum ProcessExitStatus: Sendable, Equatable {
+@_spi(Internals)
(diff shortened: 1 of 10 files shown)


Sources/ScipioKit/Executor.swift
```diff
@@ -80,18 +86,19 @@ Unknown error occurered.
...
     private let fileSystem: any FileSystem
 
-    init(errorDecoder: Decoder = StandardErrorOutputDecoder(), fileSystem: some FileSystem = LocalFileSystem()) {
+    public init(errorDecoder: Decoder = StandardErrorOutputDecoder(), fileSystem: some FileSystem = LocalFileSystem.default) {
         self.errorDecoder = errorDecoder
         self.fileSystem = fileSystem
     }
 
-    var streamOutput: (@Sendable ([UInt8]) async -> Void)?
+    public  var streamOutput: (@Sendable ([UInt8]) async -> Void)?
```
ikesyo, line 99:
```suggestion
    public var streamOutput: (@Sendable ([UInt8]) async -> Void)?
```

giginet (approved):

ikesyo (approved):

PR #230, opened 2025-07-11
Migrate utility functions from swift-tools-support-core and removed it as a dependency

Ryu0118:
I migrated utility functions (e.g. `spm_shellEscaped()`) from swift-tools-support-core and removed it as a dependency.


ikesyo (approved):
LGTM 🧹

PR #234, opened 2025-08-01
Close FileHandle explicitly to prevent `bad file descriptor` error

Ryu0118:
## Abstract
Fix a rare “bad file descriptor” error by explicitly closing FileHandle.

## Motivation
In rare cases, ClangChecker fails with a “bad file descriptor” error when executing commands via ProcessExecutor:
https://github.com/giginet/Scipio/blob/28383cec93e01815c7666997c957b6b8efa989cc/Sources/ScipioKit/Producer/Cache/CacheSystem.swift#L25

According to Apple’s documentation, FileHandle usually doesn’t need to be closed manually:
> You don’t need to send closeFile() to this object or explicitly release the object after you have finished using it.

https://developer.apple.com/documentation/foundation/pipe/filehandleforreading

However, this behavior has been reported in the following Swift issue, where a "bad file descriptor" error can occur under specific conditions such as high concurrency or quick process termination:
https://github.com/swiftlang/swift/issues/57827

To prevent this, the FileHandle is now explicitly closed after reading process output.

### DIFF
--- a/Sources/ScipioKit/Executor.swift
+++ b/Sources/ScipioKit/Executor.swift
@@ -127,6 +127,7 @@ public struct ProcessExecutor<Decoder: ErrorDecoder>: Executor, Sendable {
         let localDecoder = errorDecoder
 
         let outputTask = Task {
+            defer { try? outputHandle.close() }
             for try await data in outputHandle.byteStream() {
                 let bytes = [UInt8](data)
                 await localStreamOutput?(bytes)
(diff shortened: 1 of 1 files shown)


Sources/ScipioKit/Executor.swift
```diff
@@ -127,6 +127,9 @@ public struct ProcessExecutor<Decoder: ErrorDecoder>: Executor, Sendable {
         let localDecoder = errorDecoder
 
         let outputTask = Task {
+            // Workaround to avoid "Bad file descriptor" error when executing processes at high concurrency.
+            // Investigate this section if command output is missing
```
ikesyo, line 131: Could you please add a link to https://github.com/swiftlang/swift/issues/57827?

giginet (approved):

ikesyo (approved):

PR #235, opened 2025-08-05
Fix issue where copyHeaderKeepingStructure doesn't work correctly under certain conditions

Ryu0118:
When includeDir contained components like "." or "..", the path was previously used without being standardized.
This caused incorrect behavior in the logic that extracts subdirectories, such as this line:
https://github.com/giginet/Scipio/blob/07c313d25fd302e07448e1f05b68969af06d7b3d/Sources/ScipioKit/Producer/PIF/FrameworkBundleAssembler.swift#L103-L109

Standardizing the fileURL resolves this issue by ensuring consistent path handling.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -23,7 +23,4 @@ trailing_comma:
   mandatory_comma: true
 type_body_length:
   warning: 400
-identifier_name:
-  max_length:
-    warning: 45
 
--- a/Tests/.swiftlint.yml
+++ b/Tests/.swiftlint.yml
@@ -4,3 +4,7 @@ disabled_rules:
   - large_tuple
   - type_body_length
   - file_length
+identifier_name:
+  max_length:
+    warning: 45
+


Tests/ScipioKitTests/RunnerTests.swift
```diff
@@ -268,6 +269,78 @@ final class RunnerTests: XCTestCase {
         }
     }
 
+    func testBuildClangPackageWithCustomModulePath() async throws {
```
ikesyo, line 272: I think we can test this in FrameworkModuleMapGeneratorTests as a more basic unit test.

.swiftlint.yml
```diff
@@ -23,3 +23,7 @@ trailing_comma:
   mandatory_comma: true
 type_body_length:
   warning: 400
+identifier_name:
+  max_length:
+    warning: 45
```
Ryu0118, line 28: This change is due to the identifier "clangPackageWithRelativePublicHeadersPath" exceeding the maximum allowed length. Its length is 42 characters.

Tests/ScipioKitTests/RunnerTests.swift
Ryu0118: Thank you for your review.
I moved the test to FrameworkModuleMapGeneratorTests in the following commit.
[af9a873](https://github.com/giginet/Scipio/pull/235/commits/af9a873caa2cabd6f60b021c0024694171f56725)

Tests/ScipioKitTests/FrameworkModuleMapGeneratorTests.swift
```diff
@@ -67,7 +72,29 @@ framework module MyTarget {
         #expect(generatedModuleMapContents == expectedModuleMapContents)
     }
 
+    @Test
+    func generate_keepPublicHeadersStructure_withRelativePublicHeadersPath_is_true() async throws {
```
ikesyo, line 76:
```suggestion
    func generate_keepPublicHeadersStructure_is_true_withRelativePublicHeadersPath() async throws {
```

.swiftlint.yml
ikesyo: Default value of this is 40. The modified value is still lower than the default error value so should be okay.

https://realm.github.io/SwiftLint/identifier_name.html

image

ikesyo (approved):

ikesyo (approved):

.swiftlint.yml
Ryu0118: I moved the rule to Tests/.swiftlint.yml in 8e95b49

giginet (approved):

PR #250, opened 2025-10-23
Parallelize iterations in PackageResolver to improve performance

Ryu0118:
I updated `PackageResolver` to run concurrently instead of serially where `AsyncOperations` are used.
Since running in parallel could trigger multiple resolutions of the same target, added a cache to avoid duplicate work.

Here is the benchmark:
Before: 28.7s
After: 8.8s

About 3.3× faster on average

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -39,7 +39,7 @@ let package = Package(
         .package(url: "https://github.com/giginet/PackageManifestKit",
                  from: "0.1.0"),
         .package(url: "https://github.com/mtj0928/swift-async-operations.git",
-                 from: "0.3.0"),
+                 from: "0.4.0"),
     ],
     targets: [
         .executableTarget(


giginet (commented):
Could you update swift-async-operations?

Older versions contain potential asynchronous bugs; it is advisable to explicitly update.

Sources/ScipioKit/Resolver/PackageResolver.swift
```diff
@@ -205,7 +207,7 @@ actor PackageResolver {
             modules: try await product.targets
                 .compactMap { targetName in manifest.targets.first(where: { $0.name == targetName }) }
                 .filter { shouldBuild($0.type) }
-                .asyncMap { try await self.resolve(target: $0, in: manifest) },
+                .asyncMap(numberOfConcurrentTasks: numberOfConcurrentTasks) { try await self.resolve(target: $0, in: manifest) },
```
giginet: We discussed the issue of nested parallel execution.

While there are concerns, I believe it won't be a problem in normal Swift Concurrency, so it's sufficient to just keep it in mind.

PR #251, opened 2025-11-13
Integrate `scipio-cache-storage` into Scipio package

Ryu0118:
## Motivation
[scipio-cache-storage](https://github.com/giginet/scipio-cache-storage) was originally separated because `ScipioKit` depended on `SwiftPM`, which didn’t follow semver and caused versioning issues across custom storage plugins.
Now that `ScipioKit` no longer depends on SwiftPM, this separation is unnecessary.

## Changes
This PR integrates `scipio-cache-storage` into Scipio to simplify dependencies and maintenance.

I also introduced two new modules, CacheStorage (from `scipio-cache-storage`) and ScipioKitCore.
`ScipioKitCore` contains shared models that were previously defined in ScipioKit but will be used by CacheStorage.

### DIFF
--- a/Sources/ScipioKitCore/ResolvedModels.swift
+++ b/Sources/ScipioKitCore/ResolvedModels.swift
@@ -1,27 +1,41 @@
 import Foundation
 import PackageManifestKit
 
-public struct ModulesGraph: Sendable {
-    public var rootPackage: ResolvedPackage
-    public var allPackages: [PackageID: ResolvedPackage]
-    public var allModules: Set<ResolvedModule>
+/// Represents the complete dependency graph of a package and its dependencies.
+package struct ModulesGraph: Sendable {
+    /// The root package that composes the ModulesGraph.
+    package var rootPackage: ResolvedPackage
 
-    public func package(for module: ResolvedModule) -> ResolvedPackage? {
+    /// Dictionary of all packages in the dependency graph, keyed by their unique package ID.
+    package var allPackages: [PackageID: ResolvedPackage]
+
+    /// Set of all modules across all packages in the dependency graph.
+    package var allModules: Set<ResolvedModule>
+
+    /// Returns the package that contains the specified module.
+    ///
+    /// - Parameter module: The module to look up.
+    /// - Returns: The package containing the module, or `nil` if not found.
+    package func package(for module: ResolvedModule) -> ResolvedPackage? {
         allPackages[module.packageID]
     }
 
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -75,6 +74,19 @@ let package = Package(
...
             ]
         ),
+        .target(
+            name: "CacheStorage",
+            dependencies: [
+                "ScipioKitCore"
+            ]
+        ),
+        .target(
+            name: "ScipioKitCore",
```
giginet: Implementations look good.
But I feel `ScipioKitCore` is a bit of an ambiguous name. Furthermore, it's problematic that CacheStorage depends on this. I'll think of a better name.

Sources/ScipioKitCore/ResolvedModels.swift
```diff
@@ -70,7 +76,7 @@ struct ResolvedPackage: Identifiable {
     }
 }
 
-extension ResolvedPackage {
+public extension ResolvedPackage {
     var canonicalPackageLocation: CanonicalPackageLocation {
```
Ryu0118: If we define this property as an extension on the ScipioKit side, ScipioKitCore will no longer need to have a SwiftPM directory.

Sources/ScipioKitCore/ResolvedModels.swift
```diff
@@ -127,66 +133,92 @@ struct ResolvedModule: Hashable, Sendable {
...
 
-    func recursiveModuleDependencies() throws -> [ResolvedModule] {
+    public func recursiveModuleDependencies() throws -> [ResolvedModule] {
         try topologicalSort(self.dependencies) { $0.dependencies }.compactMap { $0.module }
     }
 
-    func recursiveDependencies() throws -> [Dependency] {
+    public func recursiveDependencies() throws -> [Dependency] {
         try topologicalSort(self.dependencies) { $0.dependencies }
     }
```
Ryu0118: Similarly, if these functions and properties are defined as extensions in ScipioKit, we won’t need to bring any files other than ResolvedModels.swift into ScipioKitCore.

Sources/ScipioKitCore/ResolvedModels.swift
```diff
@@ -0,0 +1,202 @@
...
+        self.allPackages = allPackages
+        self.allModules = allModules
+    }
+}
+
+public struct PackageID: Hashable, Codable, Sendable {
+    public var description: String
+    public var packageIdentity: String
+
+    public init(
```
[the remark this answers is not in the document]
Ryu0118: And since the initializers of all structs in `ResolvedModels.swift` should also be used only within the package, we should change their access level to `package` too.

giginet (approved):

PR #253, opened 2025-11-26
Add ResolvedPackagesCacheStorage protocol to CacheStorage

Ryu0118:
Add `ResolvedPackagesCacheStorage` protocol to support caching resolved packages.


Sources/CacheStorage/CacheStorage.swift
```diff
@@ -8,13 +8,22 @@ private let jsonEncoder = {
     return encoder
 }()
 
-public protocol CacheStorage: Sendable {
+public protocol FrameworkCacheStorage: Sendable {
```
giginet: Why does it change the name?

PR #254, opened 2025-11-26
Introduce cache system for PackageResolver to improve performance

Ryu0118:
I introduce caching system for `PackageResolver` to improve performance by caching resolved package

## Motivation
Previously, Scipio depended on SwiftPM and used its APIs directly for package resolution, which was efficient. However, after #208 , Scipio introduced its own package resolution logic independent of SwiftPM. This new approach executes `swift package resolve` (equivalent to the previous resolution) and `swift package dump-package` for every checked-out repository, which introduced performance issues.

To address this, we cache the resolved packages (array of `ResolvedPackage`) using the `originHash` from `Package.resolved` as the cache key.

## Changes
  - Implement `LocalDiskCacheStorage` conforming to `ResolvedPackagesCacheStorage`
  - Add `PackageResolver.CacheSystem` to manage caching and restoration across multiple storage backends according to the policy
  - Integrate `PackageResolver.CacheSystem` into `PackageResolver` to store and retrieve resolved package data.
  - Rename `CacheStorage` to `FrameworkCacheStorage` with deprecation alias
  - Add unit tests for `PackageResolver.CacheSystem`

### DIFF
--- a/Sources/scipio/PrepareCommands.swift
+++ b/Sources/scipio/PrepareCommands.swift
@@ -45,7 +45,7 @@ extension Scipio {
             LoggingSystem.bootstrap(logLevel: logLevel)
 
             // Show deprecation warning if cachePolicy is used
-            if cachePolicy != nil  {
+            if cachePolicy != nil {
                 logger.warning("--cache-policy is deprecated. Please use --framework-cache-policy instead.")
             }
 


giginet (commented):
It seems almost too good to be true 🌸

I noticed a couple of minor points regarding CLI behavior.
First, breaking backward compatibility with `--cache-policy`. I think it's better not to change the behavior of existing options.
How about maintaining equivalent behavior while issuing a deprecation error and encouraging the use of `--framework-cache-policy`?

The second point is the default behavior of `—resolved-packages-cache-policy`.
Currently, it's nullable and falls back to `—cache-policy` if unset. I think it would be less confusing to either opt-out to local disk cache or make it experimental opt-out.

Sources/scipio/PrepareCommands.swift
```diff
@@ -45,6 +44,11 @@ extension Scipio {
             let logLevel: Logger.Level = globalOptions.verbose ? .trace : .info
             LoggingSystem.bootstrap(logLevel: logLevel)
 
+            // Show deprecation warning if cachePolicy is used
+            if let cachePolicy = cachePolicy {
```
giginet:
```suggestion
            if let cachePolicy {
```

Sources/scipio/PrepareCommands.swift
```diff
@@ -20,9 +20,22 @@ extension Scipio {
...
+                      Specify how to reuse framework cache. (\(CachePolicy.allCases.map(\.rawValue).joined(separator: " / ")))
+                      """,
                 completion: .list(CachePolicy.allCases.map(\.rawValue)))
-        var cachePolicy: CachePolicy = .project
+        var cachePolicy: CachePolicy?
+
+        @Option(name: [.customLong("framework-cache-policy")],
+                help: "Specify how to reuse framework cache. (\(CachePolicy.allCases.map(\.rawValue).joined(separator: " / ")))",
+                completion: .list(CachePolicy.allCases.map(\.rawValue)))
+        var frameworkCachePolicy: CachePolicy?
```
giginet: Set default value and avoiding to be null type.

Sources/scipio/PrepareCommands.swift
```diff
@@ -20,9 +20,22 @@ extension Scipio {
...
+
+        @Option(name: [.customLong("resolved-packages-cache-policy")],
+                help: "Specify how to reuse resolved packages cache. (\(CachePolicy.allCases.map(\.rawValue).joined(separator: " / ")))",
+                completion: .list(CachePolicy.allCases.map(\.rawValue)))
+        var resolvedPackagesCachePolicy: CachePolicy?
```
giginet: ditto

Ryu0118: We consider resolvedPackagesCachePolicy to be necessary only for large projects, so the default is intentionally set to nil.

Ryu0118: Oh sorry, since CachePolicy has a `.disabled` case, it seems that setting the default value (`.disabled`) was the correct choice instead of nil.

Sources/scipio/PrepareCommands.swift
```diff
@@ -31,18 +44,36 @@ extension Scipio {
             let logLevel: Logger.Level = globalOptions.verbose ? .trace : .info
             LoggingSystem.bootstrap(logLevel: logLevel)
 
-            let runnerCachePolicies: [Runner.Options.CachePolicy]
-            switch cachePolicy {
+            // Show deprecation warning if cachePolicy is used
+            if cachePolicy != nil {
+                logger.warning("--cache-policy is deprecated. Please use --framework-cache-policy instead.")
```
giginet: You must either forcefully override `--framework-cache-policy` or prevent it from running with an error, otherwise unintended behavior may occur.

We might not need to provide backward compatibility mode anymore 🤔
(Simply removing `-cache-policy` causes swift-argument-parser to throw an error)

giginet (approved):

PR #264, opened 2026-03-17
Propagate underlying error in Runner.Error.invalidPackage

giginet:
## Summary

- `Runner.Error.invalidPackage` now carries the underlying `Swift.Error` that caused the failure
- When `DescriptionPackage` initialization fails, the original error (e.g. package resolution failure, manifest parsing error) is preserved and displayed to the user instead of being silently discarded
- This makes it much easier to diagnose CI failures like "Failed to parse package manifests"

## Test plan

- [x] `swift build` succeeds
- [ ] Verify that when `DescriptionPackage` initialization fails, the error message now includes the underlying cause

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/ScipioKit/Runner.swift
+++ b/Sources/ScipioKit/Runner.swift
@@ -13,7 +13,7 @@ public struct Runner {
     }
 
     public enum Error: Swift.Error, LocalizedError {
-        case invalidPackage(URL)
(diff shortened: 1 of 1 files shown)


ikesyo (approved):

ikesyo:
Let's merge 🚀

PR #271, opened 2026-05-13
Fix stale dependent framework caches by propagating dependency cache keys

### LINKED ISSUE #265: Cache key should account for dependency build state to prevent stale cache usage
## Problem

Currently, `SwiftPMCacheKey` only considers the target's own package pin state, build options, compiler version, Xcode version, and Scipio version when computing cache keys. It does **not** account for the build state of the target's dependencies.

This means that when a dependency (e.g., `FrameworkA`) is updated with a non-binary-compatible change, a downstream target (e.g., `FrameworkB` that depends on `FrameworkA`) may still have a valid cache key — even though `FrameworkB` was built against the old version of `FrameworkA`. This results in Scipio serving a stale cached binary for `FrameworkB`, which can cause link-time or runtime failures due to binary incompatibility.

## Current Cache Key Structure

From [`Sources/ScipioKit/Producer/Cache/CacheSystem.swift` (L39-L48)](https://github.com/giginet/Scipio/blob/be70991260c01b3c9a8b900430eba38e78d2ba8b/Sources/ScipioKit/Producer/Cache/CacheSystem.swift#L39-L48):

```swift
public struct SwiftPMCacheKey: CacheKey {
    public var localPackageCanonicalLocation: String?
    public var pin: Pin.State
    public var targetName: String
    var buildOptions: BuildOptions
    public var clangVersion: String
    public var xcodeVersion: XcodeVersion
    public var scipioVersion: String?
}
```

None of these fields capture the state of the target's transitive dependencies.

## Expected Behavior

When a target's dependency cache is invalidated (e.g., due to a version bump or rebuild), the cache for the dependent target should also be invalidated automatically. The cache key for a target should incorporate information about how its dependencies were built, such as:

- The cache keys (or hashes thereof) of all direct dependencies
- Whether each dependency was built as dynamic or static framework

## Reproduction Scenario

1. `FrameworkB` depends on `FrameworkA` v1.0.0
2. Scipio builds and caches both frameworks
3. `FrameworkA` is updated to v1.1.0 with non-binary-compatible API changes
4. Scipio rebuilds `FrameworkA` (its pin changed, so cache is invalidated)
5. Scipio **reuses the cached** `FrameworkB` (its own pin hasn't changed)
6. `FrameworkB` is now linked against the old `FrameworkA` binary, causing failures

## Proposed Solution

Include dependency build information in the cache key computation. One approach:

- Compute cache keys in dependency order (leaf dependencies first)
- Include each direct dependency's cache key hash as part of the dependent target's cache key
- This creates a cascade where any change in a dependency automatically invalidates all dependents

ikesyo:
## Summary
This PR fixes stale cache reuse for dependent frameworks by including dependency build state in cache keys.

Previously, a downstream target could keep reusing a cached artifact even after a dependency was rebuilt with incompatible changes.

Related issue: #265

## What Changed
- Added dependencyCacheKeyChecksums to SwiftPMCacheKey.
- Added graph-based cache key calculation (calculateCacheKeys(for:)) that computes keys in dependency order and embeds direct dependency key checksums.
- Updated framework cache flows to reuse a single cache-key map across existing framework validation, restore checks, cache writes, version file generation, and restored cache sharing.
- Preserved backward compatibility for old version files by decoding missing dependencyCacheKeyChecksums as an empty list.
- Stabilized dependency checksum ordering with a deterministic tie-breaker (targetName, then checksum).
- Kept historical behavior when cache policies are empty (cache feature is fully off).
- Fall back to building without cache participation when cache key precomputation fails, so missing local repository metadata does not abort the build before the intended build path.

## Tests
- Extended CacheSystemTests to cover encoding with dependency checksums, backward-compatible decode, direct dependency checksum propagation, and dependent key invalidation when dependency build state changes.
- Updated cache key tests to copy the PartialCacheTestPackage fixture into a temporary tagged git repository, keeping them stable in CI merge checkouts.
- FrameworkProducerTests passes locally.
- CI is passing.

## Impact
- Dependency changes now correctly invalidate dependent cache keys.
- Reduces risk of link/runtime failures caused by stale downstream cache artifacts.
- Maintains existing non-cache behavior for packages whose cache keys cannot be resolved.

### DIFF
--- a/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
+++ b/Sources/ScipioKit/Producer/Cache/CacheSystem.swift
@@ -148,6 +148,11 @@ struct CacheSystem: Sendable {
         var dependencyCacheKeyChecksums: [DependencyCacheKeyChecksum]
     }
 
+    private enum CacheKeyCalculationResult: Sendable {
+        case calculated(CacheTarget, SwiftPMCacheKey)
+        case revisionNotDetected(CacheTarget, String)
+    }
+
     enum Error: LocalizedError {
         case revisionNotDetected(String)
         case compilerVersionNotDetected
@@ -281,6 +286,8 @@ struct CacheSystem: Sendable {
     func calculateCacheKeys(for graph: DependencyGraph<CacheTarget>) async throws -> [CacheTarget: SwiftPMCacheKey] {
         let environment = try await cacheKeyEnvironment()
         var cacheKeys: [CacheTarget: SwiftPMCacheKey] = [:]
+        var unavailableTargets: Set<CacheTarget> = []
+        var reportedUnavailablePackages: Set<String> = []
         var remainingTargets = Set(graph.allNodes.map(\.value))
         var layer = graph.leafs
 
@@ -290,7 +297,19 @@ struct CacheSystem: Sendable {
                 break
             }
 
-            let inputs = try nodes.map { node in
+            var inputs: [CacheKeyCalculationInput] = []
+            var unavailablePackages: Set<String> = []
+            for node in nodes {
+                if node.children.contains(where: { unavailableTargets.contains($0.value) }) {
+                    unavailableTargets.insert(node.value)
+                    remainingTargets.remove(node.value)
+                    continue
+                }
+
+                guard node.children.allSatisfy({ cacheKeys[$0.value] != nil }) else {
+                    continue
+                }
+
                 let dependencyChecksums = try node.children.map { child in
                     let childCacheKey = cacheKeys[child.value]!
                     return DependencyCacheKeyChecksum(
@@ -298,24 +317,51 @@ struct CacheSystem: Sendable {
                         checksum: try childCacheKey.calculateChecksum()
                     )
                 }
-                return CacheKeyCalculationInput(
-                    target: node.value,
-                    dependencyCacheKeyChecksums: dependencyChecksums
+                inputs.append(
+                    CacheKeyCalculationInput(
+                        target: node.value,
+                        dependencyCacheKeyChecksums: dependencyChecksums
+                    )
                 )
-            }
-            let layerCacheKeys = try await inputs.asyncMap(numberOfConcurrentTasks: UInt(inputs.count)) { input in
-                let cacheKey = try await calculateCacheKey(
-                    of: input.target,
-                    dependencyCacheKeyChecksums: input.dependencyCacheKeyChecksums,
-                    environment: environment
-                )
-                return (input.target, cacheKey)
             }
 
-            for (target, cacheKey) in layerCacheKeys {
-                cacheKeys[target] = cacheKey
-                remainingTargets.remove(target)
+            let layerResults: [CacheKeyCalculationResult] = if inputs.isEmpty {
+                []
+            } else {
+                try await inputs.asyncMap(numberOfConcurrentTasks: UInt(inputs.count)) { input in
+                    do {
+                        let cacheKey = try await calculateCacheKey(
+                            of: input.target,
+                            dependencyCacheKeyChecksums: input.dependencyCacheKeyChecksums,
+                            environment: environment
+                        )
+                        return CacheKeyCalculationResult.calculated(input.target, cacheKey)
+                    } catch Error.revisionNotDetected(let packageName) {
+                        return CacheKeyCalculationResult.revisionNotDetected(input.target, packageName)
+                    }
+                }
             }
+
+            for result in layerResults {
+                switch result {
+                case .calculated(let target, let cacheKey):
+                    cacheKeys[target] = cacheKey
+                    remainingTargets.remove(target)
+                case .revisionNotDetected(let target, let packageName):
+                    unavailableTargets.insert(target)
+                    remainingTargets.remove(target)
+                    unavailablePackages.insert(packageName)
+                }
+            }
+
+            for packageName in unavailablePackages.subtracting(reportedUnavailablePackages).sorted() {
+                logger.warning(
+                    // swiftlint:disable:next line_length
+                    "⚠️ Cache key is unavailable because \(packageName) has no revision. Skip cache participation for this package and its dependents.",
+                    metadata: .color(.yellow)
+                )
+            }
+            reportedUnavailablePackages.formUnion(unavailablePackages)
 
             let parentNodes = nodes.flatMap { node in
                 node.parents.compactMap(\.reference)
@@ -324,7 +370,10 @@ struct CacheSystem: Sendable {
             layer = parentNodes.filter { node in
                 remainingTargets.contains(node.value)
                     && seenTargets.insert(node.value).inserted
-                    && node.children.allSatisfy { cacheKeys[$0.value] != nil }
+                    && (
+                        node.children.contains(where: { unavailableTargets.contains($0.value) })
+                            || node.children.allSatisfy { cacheKeys[$0.value] != nil }
+                    )
             }
         }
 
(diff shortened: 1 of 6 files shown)


giginet:
Thanks for the fix — the propagation approach via `dependencyCacheKeyChecksums` looks right and the backward-compatible decode is well handled. Two concerns I'd like to raise before merging.

## 1. `precomputeCacheKeysIfPossible` swallows all errors silently

```swift
} catch {
    logger.warning(
        "⚠️ Cache key precomputation failed. Continue build without cache restore/save for this run.",
        metadata: .color(.yellow)
    )
    return [:]
}
```

Previously, errors from `calculateCacheKey(of:)` such as `revisionNotDetected`, `compilerVersionNotDetected`, and `xcodeVersionNotDetected` would propagate and abort the build. With this change, **any** error falls back silently to a "cache fully disabled" mode for the whole run. Two problems:

- The warning does not include the underlying error, so a user running `prepare` mode on a large dependency tree has no way to discover why caching was silently disabled. They will just observe slow rebuilds.
- The catch is too broad. If the intent is to tolerate missing git metadata in fixtures (as the PR description says), only `Error.revisionNotDetected` should warrant the fallback. `compilerVersionNotDetected` / `xcodeVersionNotDetected` indicate genuinely broken environments and should still abort.

Suggested changes:
- At minimum, include `error.localizedDescription` (or `\(error)`) in the warning so users can diagnose.
- Narrow the catch to the specific case(s) that justify a fallback and rethrow the rest.
- Optionally, gate the fallback behind a CLI flag so strict cache validation can remain the default.

## 2. Cache key precomputation is now fully serialized

`calculateCacheKeys(for:)` iterates `graph.allNodes` and awaits each `calculateCacheKeyRecursively` call sequentially. Each call internally spawns subprocesses for `clang --version`, `xcodebuild -version`, `git describe`, and `git rev-parse`. Previously these were invoked inside `withTaskGroup` blocks in `validateExistingFrameworks` / `restoreCaches` / `shareCachesToStorage` and ran in parallel across targets. For projects with many dependencies this could regress wall-clock time noticeably.

Suggested changes (in order of effort/payoff):

- `clangVersion` and `xcodeVersion` are constants for the lifetime of a Scipio run. Fetch them once (e.g. at `CacheSystem` initialization or at the start of `calculateCacheKeys(for:)`) and reuse, instead of re-running the subprocesses for every target. This is cheap and recovers most of the regression.
- Parallelize same-level dependencies inside `calculateCacheKeys(for:)` while preserving topological ordering — for example, processing the graph layer by layer with `withTaskGroup`.

The first one alone is low-risk and a clear win.

ikesyo:
@giginet Thanks for the feedback. I addressed the points. Please re-review the changes.

520a4eb...df83ef2

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -147,9 +162,29 @@ struct FrameworkProducer {
...
+    ) async throws -> [CacheSystem.CacheTarget: SwiftPMCacheKey] {
+        do {
+            return try await cacheSystem.calculateCacheKeys(for: targetGraph)
+        } catch CacheSystem.Error.revisionNotDetected(let packageName) {
+            // Fall back to "build without cache participation" for targets whose cache keys
+            // cannot be resolved (for example, missing local git metadata in fixtures).
+            logger.warning(
+                "⚠️ Cache key precomputation failed because \(packageName) has no revision. Continue build without cache restore/save for this run.",
+                metadata: .color(.yellow)
+            )
```
Ckitakishi: `precomputeCacheKeysIfPossible` returns `[:]` when any package hits `revisionNotDetected`, so restore / save / version files are all disabled for the whole run. Can we keep a partial map and only skip the failing package and its dependents?

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -107,7 +205,8 @@ struct CacheSystem: Sendable {
                                 "🚀 Cache \(frameworkName) to cache storage: \(storageName)",
                                 metadata: .color(.green)
                             )
-                            try await cacheFramework(target, at: frameworkPath, to: storage)
+                            guard let cacheKey = cacheKeys[target] else { return }
```
Ckitakishi: The "🚀 Cache ..." log comes before the `cacheKeys[target]` check. When the key is missing, it looks like a success in the log but nothing happens. Can we move the guard above the log, or add a warning in the nil branch?

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -373,16 +415,25 @@ struct FrameworkProducer {
...
 
-    private func generateVersionFile(for target: CacheSystem.CacheTarget, using cacheSystem: CacheSystem) async {
+    private func generateVersionFile(
+        for target: CacheSystem.CacheTarget,
+        using cacheSystem: CacheSystem,
+        cacheKeys: [CacheSystem.CacheTarget: SwiftPMCacheKey]
+    ) async {
         do {
-            try await cacheSystem.generateVersionFile(for: target)
+            guard let cacheKey = cacheKeys[target] else { return }
```
Ckitakishi: `generateVersionFile` returns silently when `cacheKeys[target]` is missing. Can we keep the existing "Could not create VersionFile" warning here too?

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -171,6 +270,72 @@ struct CacheSystem: Sendable {
     }
 
     func calculateCacheKey(of target: CacheTarget) async throws -> SwiftPMCacheKey {
```
Ckitakishi: `calculateCacheKey(of:)` and the keyless `generateVersionFile(for:)` still produce keys with empty `dependencyCacheKeyChecksums`, so they never match the graph-computed keys for non-leaf targets. Only tests use them now, but it's easy to misuse later. Can we limit these APIs or route callers through the graph-based calculation?

Sources/ScipioKit/Producer/Cache/CacheSystem.swift
```diff
@@ -171,6 +270,72 @@ struct CacheSystem: Sendable {
...
+                        targetName: child.value.buildProduct.target.name,
+                        checksum: try childCacheKey.calculateChecksum()
+                    )
+                }
+                return CacheKeyCalculationInput(
+                    target: node.value,
+                    dependencyCacheKeyChecksums: dependencyChecksums
+                )
+            }
+            let layerCacheKeys = try await inputs.asyncMap(numberOfConcurrentTasks: UInt(inputs.count)) { input in
```
Ckitakishi: `asyncMap(numberOfConcurrentTasks: UInt(inputs.count))` has no upper limit per layer, and each key calculation may run git. Can we cap it at `CacheSystem.defaultParallelNumber`?

Sources/ScipioKit/Producer/FrameworkProducer.swift
```diff
@@ -147,9 +162,29 @@ struct FrameworkProducer {
...
+            return try await cacheSystem.calculateCacheKeys(for: targetGraph)
+        } catch CacheSystem.Error.revisionNotDetected(let packageName) {
+            // Fall back to "build without cache participation" for targets whose cache keys
+            // cannot be resolved (for example, missing local git metadata in fixtures).
+            logger.warning(
+                "⚠️ Cache key precomputation failed because \(packageName) has no revision. Continue build without cache restore/save for this run.",
+                metadata: .color(.yellow)
+            )
+            return [:]
+        } catch {
```
Ckitakishi: `catch { throw error }` is redundant and can be removed.

Ckitakishi (approved):
LGTM, thank you!

PR #273, opened 2026-05-22
Fix process executor hang on inherited pipes

### LINKED ISSUE #272: v0.32.1 will not print any log and return process result on GitHubActions.
Hi

When I tried to update latest v0.32.1 from v0.29.1, scipio will not print any log (even the first log of `🔁 Resolving Dependencies...`) and not exit the process.
Resultly github actions default 6h timeouted on our CI.

Image

Investigate and fix this issue by ClaudeCode https://github.com/r-plus/Scipio/commit/3c8cf970857f4ed1dfca16caf9b9a30d991e54a2 is resolved this issue.

Can you make appropriate corrections?

---

<details>
<summary>Investigation result by ClaudeCode</summary>

```
  Symptoms

  - On GitHub Actions (macOS runner), scipio produces no stdout/stderr output during a build.
  - The internal work appears to make progress.
  - After the work completes, the process fails to terminate and waits until the 6-hour job timeout.
  - The same scipio invocation runs to completion normally on a local macOS machine.

  Both regressions appeared with the switch from TSCBasic.Process to Foundation.Process in 0.30.0.

  Root causes

  There are two independent issues that happen to surface together in CI.

  Issue 1 — Parent process stdout is block-buffered when not a TTY

  Sources/ScipioKit/Logger.swift writes log lines via Swift's print(), which goes through the C runtime's
  stdout. The C runtime selects a buffering mode at startup based on the descriptor:

  - TTY → line-buffered (flushed on \n)
  - Pipe / file → block-buffered (flushed only when the buffer fills or the process exits)

  Locally, scipio runs attached to a terminal, so log lines like 🔁 Resolving Dependencies... are flushed
  immediately. On GHA, the runner captures the step's output via a pipe, so scipio's stdout is
  block-buffered and short log lines are withheld until process exit.

  This is purely a parent-side buffering issue and is unrelated to the subprocess plumbing.

  Issue 2 — Hang because pipe EOF never arrives

  Sources/ScipioKit/Executor.swift (the new Foundation.Process-based ProcessExecutor) reads child
  stdout/stderr from a Pipe() via FileHandle.readabilityHandler, exposed as an AsyncThrowingStream
  (byteStream()). The execute flow synchronously called process.waitUntilExit() and then await
  outputTask.value / await errorOutputTask.value. The async stream completes only when readabilityHandler
  is invoked with empty data (EOF).

  This breaks down for xcodebuild / Swift Build:

  1. xcodebuild launches XCBBuildService (the Swift Build daemon) if one is not already running.
  2. When scipio is the process that triggers this on a fresh machine, XCBBuildService is spawned as a
  descendant of scipio and inherits the stdout/stderr pipe FDs.
  3. XCBBuildService is a long-lived daemon. Even after xcodebuild itself exits, the daemon keeps the
  write end of the pipe open.
  4. The read end therefore never observes EOF → readabilityHandler is never called with empty data → the
  AsyncThrowingStream never finishes → outputTask.value never returns → the
  withCheckedThrowingContinuation never resumes.

  Result: a permanent hang until the job timeout.

  Additionally, the previous version of the executor set process.terminationHandler after
  process.waitUntilExit() had returned. Foundation does not invoke a terminationHandler that is installed
  after the process is already terminated, so even if EOF were reached the continuation could not be
  resumed reliably.

  Why this reproduces only on CI

  - Locally, developers normally have Xcode.app running at some point, which has already started
  XCBBuildService with its own stdio. Subsequent xcodebuild invocations attach to the existing daemon via
  XPC and do not inherit scipio's pipe FDs. The pipe closes cleanly when xcodebuild exits, EOF arrives,
  and the hang does not occur.
  - GHA's macOS runner starts clean every job. The first xcodebuild call spawns XCBBuildService as a
  descendant of scipio, the FD inheritance occurs, and the hang manifests.

  The block-buffering issue (Issue 1) is also TTY-vs-pipe specific and naturally appears only in CI.

  TSCBasic.Process used DispatchIO for reads and signalled completion through a separate path, so it did
  not depend on pipe EOF to know the process had finished. That is why 0.30.0 (the move to
  Foundation.Process) introduced these regressions.

  Fix

  a. Force line buffering in the scipio entry point

  In Sources/scipio/Scipio.swift, override static func main() and call:

  setvbuf(stdout, nil, _IOLBF, 0)
  setvbuf(stderr, nil, _IOLBF, 0)

  before dispatching to the parsed command. This makes print()-based log output flush on every line
  regardless of whether stdout is a pipe, so GHA shows progress in real time.

  b. Make ProcessExecutor independent of pipe EOF

  In Sources/ScipioKit/Executor.swift:

  1. Install process.terminationHandler before calling process.run() and remove the synchronous
  process.waitUntilExit(). Completion is signalled through the termination handler only.
  2. Replace the actor-based ProcessOutputBuffer with a lock-based, Sendable-checked byte buffer so it can
   be read synchronously from readabilityHandler and terminationHandler.
  3. Read stdout/stderr directly inside readabilityHandler: append to the lock-based buffer; for stdout,
  also forward each chunk to streamOutput through a small serial Task-chain (StreamDeliverer) so the
  XCBuild JSON parser sees chunks in order.
  4. In terminationHandler, do not wait for EOF. Detach the readability handlers, close the read handles,
  snapshot the buffers, and resume the continuation. Trailing bytes that a stray grandchild daemon might
  write after the immediate child has exited are intentionally discarded — they are not part of the
  build's logical output.

  With this, the executor completes as soon as the immediate child process exits, even if a descendant
  daemon keeps the pipe's write end open. The previously observed 6-hour hang is no longer possible, and
  streamOutput continues to deliver xcbuild messages live during the build.

  Summary of behaviour after the fix

  ┌────────────────────────────────────────┬─────────────────┬───────────────────────────────────────┐
  │                Concern                 │     Before      │                 After                 │
  ├────────────────────────────────────────┼─────────────────┼───────────────────────────────────────┤
  │ 🔁 Resolving Dependencies... visible   │ No              │ Yes (line-buffered stdout)            │
  │ in CI                                  │                 │                                       │
  ├────────────────────────────────────────┼─────────────────┼───────────────────────────────────────┤
  │ Real-time xcbuild log in CI            │ No              │ Yes (streamOutput via ordered         │
  │                                        │                 │ delivery)                             │
  ├────────────────────────────────────────┼─────────────────┼───────────────────────────────────────┤
  │ Process hangs at end of build          │ Yes (6h         │ No (termination is not gated on pipe  │
  │                                        │ timeout)        │ EOF)                                  │
  ├────────────────────────────────────────┼─────────────────┼───────────────────────────────────────┤
  │ Local macOS behaviour                  │ OK              │ OK (unchanged)                        │
  └────────────────────────────────────────┴─────────────────┴───────────────────────────────────────┘
```

</details>

ikesyo:
## Summary

Fixes the process execution hang reported in #272 when a child process exits but a long-lived descendant keeps stdout/stderr pipe file descriptors open.

## Changes

- Make `ProcessExecutor` complete based on the immediate child process termination instead of waiting for pipe EOF.
- Introduce `ProcessOutputReader`, which owns each non-blocking `FileHandle` and manages its readability notifications and closure.
- Serialize read, append, and close operations through a single locked state so queued readability callbacks cannot access a closed or reused file descriptor.
- Track EOF explicitly, stop readability notifications when observation finishes, and perform one final non-blocking drain at process termination.
- Preserve ordered `streamOutput` delivery and wait for all scheduled callbacks before returning from `execute`.
- Retry interrupted `fcntl` calls and fail before starting the process if a pipe cannot be configured for non-blocking reads.
- Flush Scipio log output after each line so CI pipe output appears promptly.
- Add regression coverage for inherited pipe handles, stream delivery completion, EOF handling, and stale drains after closure.

## Root Cause

The Foundation.Process-based executor waited for stdout/stderr streams to reach EOF after `waitUntilExit()`. On fresh CI environments, build tools can spawn long-lived descendants that inherit the write end of those pipes, so EOF may never arrive even after the immediate child exits.

Completing at immediate-child termination requires a final drain without waiting for EOF. The reader therefore uses non-blocking descriptors and synchronizes draining with closure so a queued readability callback cannot race with descriptor closure or reuse.

## Validation

- `swift test --filter ProcessOutputReaderTests`
- `swift test --filter ProcessExecutorTests`
- `swift test`
- Confirmed by @r-plus https://github.com/giginet/Scipio/issues/272#issuecomment-4524278316

### DIFF
--- a/Sources/ScipioKit/Executor.swift
+++ b/Sources/ScipioKit/Executor.swift
@@ -1,11 +1,64 @@
 import Foundation
+import Darwin
+import os
 
-/// Buffer to collect process output bytes.
-private actor ProcessOutputBuffer {
-    private(set) var bytes: [UInt8] = []
+/// Buffer to collect process output bytes from Foundation callbacks.
+private final class ProcessOutputBuffer: Sendable {
+    private let storage = OSAllocatedUnfairLock(initialState: [UInt8]())
+    private let lastStreamOutputTask = OSAllocatedUnfairLock(initialState: Task<Void, Never> {})
+    private let streamOutput: (@Sendable ([UInt8]) async -> Void)?
 
-    func append(_ newBytes: [UInt8]) {
-        bytes += newBytes
+    init(streamOutput: (@Sendable ([UInt8]) async -> Void)? = nil) {
+        self.streamOutput = streamOutput
+    }
+
+    func drain(from fileDescriptor: Int32) {
+        for bytes in Self.drainAvailableBytes(from: fileDescriptor) {
+            storage.withLock {
+                $0.append(contentsOf: bytes)
+            }
+            if let streamOutput {
+                lastStreamOutputTask.withLock {
+                    let previousTask = $0
+                    $0 = Task {
+                        await previousTask.value
+                        await streamOutput(bytes)
+                    }
+                }
+            }
+        }
+    }
+
+    var snapshot: [UInt8] {
+        storage.withLock { $0 }
+    }
+
+    func finishStreaming() async {
+        let task = lastStreamOutputTask.withLock { $0 }
+        await task.value
+    }
+
+    private static func drainAvailableBytes(from fileDescriptor: Int32) -> [[UInt8]] {
+        var chunks: [[UInt8]] = []
+        var buffer = [UInt8](repeating: 0, count: 4096)
+
+        while true {
+            let byteCount = buffer.withUnsafeMutableBytes {
+                Darwin.read(fileDescriptor, $0.baseAddress, $0.count)
+            }
+
+            switch byteCount {
+            case let byteCount where byteCount > 0:
+                chunks.append(Array(buffer.prefix(byteCount)))
+            case 0:
+                return chunks
+            default:
+                if errno == EINTR {
+                    continue
+                }
+                return chunks
+            }
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


Sources/ScipioKit/Executor.swift
```diff
@@ -141,81 +204,74 @@ public struct ProcessExecutor<Decoder: ErrorDecoder>: Executor, Sendable {
...
-                )
-
-                switch exitStatus {
-                case .terminated(let code) where code == 0:
-                    continuation.resume(returning: result)
-                case .terminated:
-                    do {
-                        let errorOutput = try localDecoder.decode(result)
+                Task {
+                    outputReader.drain(from: outputFileDescriptor)
```
Ckitakishi: `readabilityHandler = nil` does not wait for an in-flight handler, so the final drain/close can still race with a readability drain on the same fd. That can reorder output, miss bytes in the final snapshot, or let a stale handler touch a reused fd after close. Could we make the reader own the `FileHandle` and guard read/append/close with one closed-state lock? A small concurrency smoke test would help.

Sources/ScipioKit/Executor.swift
```diff
@@ -1,11 +1,74 @@
...
+            let byteCount = buffer.withUnsafeMutableBytes {
+                Darwin.read(fileDescriptor, $0.baseAddress, $0.count)
+            }
 
-    func append(_ newBytes: [UInt8]) {
-        bytes += newBytes
+            switch byteCount {
+            case let byteCount where byteCount > 0:
+                chunks.append(Array(buffer.prefix(byteCount)))
+            case 0:
```
Ckitakishi, line 136: `read` returning 0 means EOF. Can we mark the reader closed and remove the handler here so EOF does not keep retriggering readability?

ikesyo: Addressed in 7ab41f2172bf9758a10f1ce838b02defbf69de9a

Sources/ScipioKit/Executor.swift
ikesyo: Addressed in 7ab41f2172bf9758a10f1ce838b02defbf69de9a

Ckitakishi (approved):
LGTM, thank you!

PR #276, opened 2026-07-08
Avoid exponential resolved packages cache expansion

Ckitakishi:
## Summary

The resolved packages cache encodes `[ResolvedPackage]` directly. Since `ResolvedModule` and `ResolvedProduct` embed their dependencies by value, encoding walks the shared dependency DAG as a tree, and the work doubles at every shared level:

```
graph (linear size)         encoded as a tree (doubles per level)

A ──▶ B ──▶ C                        A
│     ▲                             / \
└──▶ P┘                            B   P
(product P wraps B)                │   │
                                   C   B
                                       │
                                       C
```

On large graphs this means huge `ResolvedPackages_*.json` files, and in the worst case an OOM during the cache write. This PR stores a flat snapshot instead: each package, module, and product appears once, and dependency edges are table indices.

```
modules:  [A, B, C]        products: [P]
edges:    A → module B, A → product P, P → module B, B → module C
```

`ResolvedModule` and `ResolvedProduct` hashing had the same recursive traversal problem, so this PR also makes their equality and hashing identity-based.

## Changes

- Add `ResolvedPackagesSnapshot`, the flat format. Restoring validates the file (version, references, duplicate identities, cycles, depth) and treats anything inconsistent as a cache miss. A fixture test pins the version 1 bytes, so a format change without a `currentFormatVersion` bump fails in CI.
- Store it from `PackageResolver.LocalDiskCacheStorage`.
- Make `LocalDiskCacheStorage.existsValidCache` validate that the cache file actually restores instead of only checking existence. Without this, the cache-sharing step would skip storages holding a legacy or corrupted file, leaving them broken for as long as an earlier storage keeps hitting.
- Make `ResolvedModule`/`ResolvedProduct` equality and hashing identity-based (`packageID` + name). The resolver creates at most one value per identity, so identity equality is an equivalence; the invariant is documented on `ResolvedModule.==`.

## Behavior Change: `ResolvedModule` / `ResolvedProduct` Equality

This is a deliberate semantic change to public types: `==` and `hash(into:)` now compare identity (`packageID` + name) only; `dependencies`, `type`, `localPackageURL`, and other structural content no longer participate. Within Scipio's construction paths this is an equivalence: the resolver and the snapshot restorer create at most one value per identity. Values decoded directly or mutated through the public properties are outside that invariant; code that needs structural comparison should compare fields explicitly.

## Migration Note

Old cache files encoded as `[ResolvedPackage]` are treated as unsupported: discarded when first read (a fetch, or the validity check during cache sharing) and regenerated after a full resolve or overwritten by the sharing step. This only affects the resolved packages metadata cache, not XCFramework artifact caches.


Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    /// A structural inconsistency found while restoring a snapshot.
+    /// Storages should treat these errors as a cache miss.
+    package enum RestoreError: LocalizedError, Equatable {
+        case unsupportedFormatVersion(found: Int, supported: Int)
+        case invalidModuleReference(index: Int)
+        case invalidProductReference(index: Int)
+        case duplicateModuleIdentity(ResolvedIdentity)
+        case duplicateProductIdentity(ResolvedIdentity)
```
S-Shimotori, line 84: ScipioKitTests doesn't check this case

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
```diff
@@ -0,0 +1,445 @@
...
+                    let (product, index) = pendingProducts[nextProduct]
+                    nextProduct += 1
+                    products[index].moduleIndices = product.modules.map { self.index(of: $0) }
+                }
+            }
+        }
+
+        /// Returns the module's table index, registering it on first visit so
+        /// that shared modules are stored only once.
+        private mutating func index(of module: ResolvedModule) -> Int {
```
S-Shimotori: Could you rename this method?

The standard `Array`'s `index(of:)` is not mutating.
https://developer.apple.com/documentation/swift/array/index(of:)
(And this method is deprecated now)

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
```diff
@@ -0,0 +1,445 @@
...
+                    localPackageURL: module.localPackageURL,
+                    packageID: module.packageID,
+                    resolvedModuleType: module.resolvedModuleType
+                )
+            )
+            pendingModules.append((module, index))
+            return index
+        }
+
+        private mutating func index(of product: ResolvedProduct) -> Int {
```
S-Shimotori: ditto

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
```diff
@@ -0,0 +1,445 @@
...
+        /// bounds checks, so a corrupt snapshot must be rejected here to
+        /// degrade to a miss instead of a crash.
+        private func validateReferences() throws {
+            for module in snapshot.modules {
+                for dependency in module.dependencies {
+                    switch dependency.kind {
+                    case .module:
+                        try validateModuleReference(dependency.index)
+                    case .product:
+                        try validateProductReference(dependency.index)
```
S-Shimotori, line 299: Maybe ScipioKitTests doesn't test this case

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
```diff
@@ -0,0 +1,445 @@
...
+                    }
+                }
+            }
+            for product in snapshot.products {
+                for index in product.moduleIndices {
+                    try validateModuleReference(index)
```
S-Shimotori, line 305: Maybe ScipioKitTests doesn't test this case

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
```diff
@@ -0,0 +1,445 @@
...
+                }
+            }
+            for package in snapshot.packages {
+                for index in package.targetIndices {
+                    try validateModuleReference(index)
```
S-Shimotori, line 310: Maybe ScipioKitTests doesn't test this case

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
```diff
@@ -0,0 +1,445 @@
...
+///
+/// The model types embed their dependencies recursively by value, so their
+/// synthesized `Codable` output duplicates every shared subtree: exponential
+/// in graph depth. The snapshot stores each package, module, and product
+/// exactly once and references them by table index, so its size stays linear
+/// in the graph size.
+package struct ResolvedPackagesSnapshot: Codable, Sendable, Equatable {
+    /// Bump this when the stored representation changes; a snapshot with a
+    /// different version fails to restore, which storages treat as a miss.
+    package static let currentFormatVersion = 1
```
S-Shimotori, line 16: I'm concerned that we may forget to bump this value.

How about

- writing definition, schema, or something
- adding a test that fails when we forget to update

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
Ckitakishi: Renamed to `registerIndex(of:)` in 9931692.

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
Ckitakishi: Added a fixture test in 8a51e16. The fixture pins the version 1 bytes: if the stored format changes without bumping `currentFormatVersion`, the test fails and writes the new bytes next to the fixture as a `.actual` file. To adopt a deliberate change, bump the version and replace the fixture with that file. The fixture also works as the format definition.

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
Ckitakishi: Covered in 9931692: the tampering tests are now one parameterized test, `rejectsTamperedSnapshot`. It also covers the other untested paths you pointed out below (the product dependency edge, `moduleIndices`, and `targetIndices`).

Sources/ScipioKitCore/ResolvedPackagesSnapshot.swift
S-Shimotori: `registerIndex(of:)` seems that this method register not an object but only its index.
I prefer `func register(_:) -> Int`

Ckitakishi: Done, renamed to register(_:) in [9d79e99](https://github.com/giginet/Scipio/pull/276/commits/9d79e99552c2fc38b25655e043f9fddd3d953474)

S-Shimotori (approved):
Thank you.
I'll leave it up to you whether you improve test coverage more or not.

Ckitakishi:
@S-Shimotori Thank you for the review! All the untested cases are covered by `rejectsTamperedSnapshot` in [9931692](https://github.com/giginet/Scipio/commit/9931692dbfedf84ac06f368a8d497b4fba9d36f2), I'll merge this PR.

PR #277, opened 2026-07-09
Strip top-level SwiftPM traits from dump-package JSON before decoding

Ckitakishi:
## Summary

`ManifestLoader` decodes `swift package dump-package` output into
`PackageManifestKit.Manifest`. Since SwiftPM 6.1, `dump-package` emits package
traits as objects:

```json
"traits" : [
  { "name" : "default", "enabledTraits" : ["FoundationURL"] },
  { "name" : "FoundationURL", "description" : "...", "enabledTraits" : [] }
]
```

but `PackageManifestKit` 0.2.0 models `Manifest.traits` as `[String]?`, so any
manifest that declares traits (e.g. swift-http-types 1.6.0) fails to decode,
aborting the build:

```
DecodingError.typeMismatch: expected String, found dictionary. Path: traits[0]
```

Scipio never reads top-level `Manifest.traits`, so this PR strips that field from
the `dump-package` JSON before decoding.

## Note

Stopgap until the root fix in PackageManifestKit (model `traits` as
`[TraitDescription]?`) ships and the dependency is bumped.

### DIFF
--- a/Sources/ScipioKit/Resolver/ManifestLoader.swift
+++ b/Sources/ScipioKit/Resolver/ManifestLoader.swift
@@ -44,8 +44,24 @@ struct ManifestLoader: @unchecked Sendable {
             throw Error.utf8EncodingFailed
         }
 
-        let manifest = try jsonDecoder.decode(Manifest.self, from: manifestData)
-        return manifest
+        return try decodeManifest(from: manifestData)
+    }
+
+    // PackageManifestKit 0.2.0 expects Manifest.traits as [String]?, but SwiftPM 6.1+
+    // dump-package emits trait objects. Since Scipio does not read top-level traits,
+    // decode directly and, only on failure, drop the field and retry. Remove once
+    // PackageManifestKit models Manifest.traits as [TraitDescription]?.
+    private func decodeManifest(from data: Data) throws -> Manifest {
+        do {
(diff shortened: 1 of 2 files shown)


Sources/ScipioKit/Resolver/ManifestLoader.swift
```diff
@@ -44,8 +44,24 @@ struct ManifestLoader: @unchecked Sendable {
...
 
-        let manifest = try jsonDecoder.decode(Manifest.self, from: manifestData)
-        return manifest
+        return try decodeManifest(from: manifestData)
+    }
+
+    // PackageManifestKit 0.2.0 expects Manifest.traits as [String]?, but SwiftPM 6.1+
+    // dump-package emits trait objects. Since Scipio does not read top-level traits,
+    // decode directly and, only on failure, drop the field and retry. Remove once
+    // PackageManifestKit models Manifest.traits as [TraitDescription]?.
```
S-Shimotori, line 53: How about creating an issue on PackageManifestKit repository?

Ckitakishi: A PR has already been created! https://github.com/giginet/PackageManifestKit/pull/3

S-Shimotori (approved):

PR #278, opened 2026-07-09
Rewrite C header includes for framework consumers

Ckitakishi:
## Summary

C packages address public headers relative to their include directory:

```c
#include <core/core.h>
```

SwiftPM resolves this form through the `-I` / `HEADER_SEARCH_PATHS` entries it injects. Consumers of the prebuilt XCFrameworks only get framework search paths (`-F`), which resolve the framework-relative form instead:

```c
#include <CoreLib/core/core.h>
```

So headers written in the SwiftPM form are unusable from the prebuilt frameworks.

This PR rewrites copied public headers from the first form to the second when an include points at the target itself or one of its visible dependencies. Replacement paths encode the header's final layout under `*.framework/Headers`, covering both `keepPublicHeadersStructure` and flattened headers.

## Behavior

- System headers, quoted includes, and paths outside the dependency closure are untouched.
- If multiple visible modules expose the same include path, Scipio logs a warning and leaves the include as-is rather than picking a module by iteration order.
- Headers with no matching includes round-trip byte-for-byte, so existing outputs are unaffected.

## Testing

- Unit tests cover the rewrite rules and table construction: nested vs flattened layout, ambiguous paths, and symlinked headers.
- An end-to-end test compiles a consumer against the produced frameworks using only `-F`, with none of the `-I` paths SwiftPM would inject, covering both cross-module and same-module includes.

### DIFF
--- a/Sources/ScipioKit/Producer/PIF/CHeaderIncludeRewriter.swift
+++ b/Sources/ScipioKit/Producer/PIF/CHeaderIncludeRewriter.swift
@@ -18,16 +18,18 @@ import ScipioKitCore
 struct CHeaderIncludeRewriter {
     /// Source include path to framework include path, e.g. `"core/core.h" -> "CoreLib/core/core.h"`.
     private let replacementsByHeaderPath: [String: String]
+    private let fileSystem: any FileSystem
 
-    init(replacementsByHeaderPath: [String: String]) {
+    init(replacementsByHeaderPath: [String: String], fileSystem: some FileSystem = LocalFileSystem.default) {
         self.replacementsByHeaderPath = replacementsByHeaderPath
+        self.fileSystem = fileSystem
     }
 
     /// Builds the table from a target and its visible dependency closure.
     init(
         modules: some Sequence<ResolvedModule>,
         keepPublicHeadersStructure: (ResolvedModule) -> Bool,
-        fileSystem: any FileSystem = LocalFileSystem.default
+        fileSystem: some FileSystem = LocalFileSystem.default
     ) {
         var table: [String: String] = [:]
         var ambiguousPaths: Set<String> = []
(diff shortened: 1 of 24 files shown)


Sources/ScipioKit/Producer/PIF/CHeaderIncludeRewriter.swift
```diff
@@ -0,0 +1,111 @@
...
+                    keepPublicHeadersStructure: keepsStructure
+                ).joined(separator: "/")
+                let replacement = "\(moduleName)/\(inFrameworkPath)"
+
+                if let existing = table[sourceRelativePath], existing != replacement {
+                    // Do not guess when two modules provide the same include path.
+                    ambiguousPaths.insert(sourceRelativePath)
+                    logger.warning(
+                        "Public header \(sourceRelativePath) is provided by multiple modules; leaving its includes unrewritten"
+                    )
```
S-Shimotori, line 63: How about adding ⚠️ and `metadata: .color(.yellow)`?

Sources/ScipioKit/Producer/PIF/CHeaderIncludeRewriter.swift
```diff
@@ -0,0 +1,111 @@
...
+        }
+        self.replacementsByHeaderPath = table
+    }
+
+    var isEmpty: Bool { replacementsByHeaderPath.isEmpty }
+
+    // Angle includes only; quoted includes keep their source-relative semantics.
+    private static let includeRegex = try! NSRegularExpression(
+        pattern: #"^(\s*#\s*(?:include|import)\s*)<([^>]+)>"#
+    )
```
S-Shimotori: Probably we can use Swift's `Regex`

Sources/ScipioKit/Producer/PIF/CHeaderIncludeRewriter.swift
Ckitakishi: Addressed at: https://github.com/giginet/Scipio/pull/278/commits/e56026f40cc8ac341303b84f0a628ad8855999c7

Sources/ScipioKit/Producer/PIF/CHeaderIncludeRewriter.swift
Ckitakishi: Good catch, converted in https://github.com/giginet/Scipio/pull/278/commits/8ccdebfaff2e1ee56e453369fedd4e0d22dcb9bc

S-Shimotori (approved):

PR #279, opened 2026-07-14
Package system-library targets as header-only frameworks

Ckitakishi:
## Summary

SwiftPM's `.systemLibrary` targets carry a module map and optional headers and have nothing to compile:

```
Sources/SysShim/
├── module.modulemap
└── shim.h              // #include <core/core.h>
```

SwiftPM exposes such modules to importers through injected search paths, which only resolve while the package sources are on disk. Prebuilt-framework consumers only get `-F`, so today the producer dies with `fatalError("Unexpected target type")`, and even if it skipped the target, `import SysShim` in consumers would fail with `missing required module 'SysShim'`.

This PR packages system-library targets as header-only frameworks, so the module resolves through the framework search paths consumers already have:

```
SysShim.xcframework/ios-arm64/SysShim.framework/
├── SysShim                    // stub binary; XCFramework creation requires one
├── Headers/shim.h             // #include <CoreLib/core/core.h>, rewritten by #278
├── Modules/module.modulemap   // framework module SysShim { ... }
└── Info.plist
```

## Behavior

- `pkgConfig` declarations log a warning instead of failing: shim-style modules still resolve through the SDK and autolinking.
- Linking dynamic frameworks no longer injects `-framework` for dependencies that produce no linkable framework: system modules resolve through the SDK, and non-producible targets build no product. Generated module maps' `link framework` sections follow the same rule.
- Targets that cannot produce a framework (executables, macros, plugins, tests) are pruned from the build-product graph together with dependencies reachable only through them. Such graphs previously crashed with `fatalError("Unexpected target type")`.
- Resolved-packages caches written before system-library support are discarded on restore.
- Discovery is through importer edges; a system-library target reachable only as a product member stays out of scope, unchanged from current behavior.

## Testing

- Unit tests cover the resolution shape, module map conversion below leading comments, cache staleness, the snapshot round-trip, and the generated Info.plist.
- An end-to-end test compiles a consumer importing the system-library module against the produced frameworks using only `-F`: the `missing required module` failure this PR removes.
- Unit tests cover the graph pruning and the linkable-dependency walk; an end-to-end test builds a dynamic framework whose library depends on an executable target. Custom module map contents are covered for system-library targets.

### DIFF
--- a/Sources/ScipioKit/DescriptionPackage.swift
+++ b/Sources/ScipioKit/DescriptionPackage.swift
@@ -128,11 +128,14 @@ private final class BuildProductsResolver {
         let targetsToBuild = try targetsToBuild()
         let products = try targetsToBuild.flatMap(resolveBuildProduct(from:))
 
-        return try DependencyGraph<BuildProduct>.resolve(
+        // Resolve the full set first so every child ID stays resolvable; `filter` reconnects
+        // parents to children, preserving the build order of the remaining products.
+        let graph = try DependencyGraph<BuildProduct>.resolve(
             Set(products),
             id: \.target.name,
             childIDs: { $0.target.dependencies.flatMap(\.moduleNames) }
         )
+        return graph.filter { $0.target.underlying.type.isFrameworkProducible }
     }
 
     private func targetsToBuild() throws -> [ResolvedModule] {
--- a/Sources/ScipioKit/DescriptionPackage.swift
+++ b/Sources/ScipioKit/DescriptionPackage.swift
@@ -72,6 +72,18 @@ extension DescriptionPackage {
     }
 }
 
+extension PackageManifestKit.Target.TargetKind {
+    /// Whether Scipio can produce an XCFramework for this target kind.
+    var isFrameworkProducible: Bool {
+        switch self {
+        case .regular, .binary, .system:
+            true
+        case .executable, .test, .plugin, .macro:
+            false
+        }
+    }
+}
+
 struct BuildProduct: Hashable, Sendable {
     var package: ResolvedPackage
     var target: ResolvedModule
--- a/Sources/ScipioKit/DescriptionPackage.swift
+++ b/Sources/ScipioKit/DescriptionPackage.swift
@@ -83,6 +83,11 @@ extension PackageManifestKit.Target.TargetKind {
         }
     }
 
+    /// Whether the produced framework is linkable: a system-library framework
+    /// is header-only and resolves through the SDK instead.
+    var isFrameworkLinkable: Bool {
+        isFrameworkProducible && self != .system
+    }
 }
 
 struct BuildProduct: Hashable, Sendable {
--- a/Sources/ScipioKit/DescriptionPackage.swift
+++ b/Sources/ScipioKit/DescriptionPackage.swift
@@ -82,6 +82,7 @@ extension PackageManifestKit.Target.TargetKind {
             false
         }
     }
+
 }
 
 struct BuildProduct: Hashable, Sendable {
@@ -126,16 +127,16 @@ private final class BuildProductsResolver {
 
     func resolveBuildProductDependencyGraph() throws -> DependencyGraph<BuildProduct> {
         let targetsToBuild = try targetsToBuild()
-        let products = try targetsToBuild.flatMap(resolveBuildProduct(from:))
+        let products = try Set(targetsToBuild.flatMap(resolveBuildProduct(from:)))
 
-        // Resolve the full set first so every child ID stays resolvable; `filter` reconnects
-        // parents to children, preserving the build order of the remaining products.
-        let graph = try DependencyGraph<BuildProduct>.resolve(
-            Set(products),
+        // Non-producible targets are pruned from the resolution with their subtrees,
+        // so edges into them are dropped here as well.
+        let productNames = Set(products.map(\.target.name))
+        return try DependencyGraph<BuildProduct>.resolve(
+            products,
             id: \.target.name,
-            childIDs: { $0.target.dependencies.flatMap(\.moduleNames) }
+            childIDs: { $0.target.dependencies.flatMap(\.moduleNames).filter(productNames.contains) }
         )
-        return graph.filter { $0.target.underlying.type.isFrameworkProducible }
     }
 
     private func targetsToBuild() throws -> [ResolvedModule] {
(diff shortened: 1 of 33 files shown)


Sources/ScipioKit/Producer/SystemLibraryPackager.swift
```diff
@@ -0,0 +1,255 @@
...
+            keepPublicHeadersStructure: true,
+            outputDirectory: frameworkOutputDir,
+            fileSystem: fileSystem,
+            headerIncludeRewriter: headerIncludeRewriter
+        )
+        try assembler.assemble()
+    }
+
+    /// The dependency closure of every module that transitively depends on this one: exactly
+    /// the compile contexts SwiftPM would have compiled these headers in.
```
S-Shimotori:
```suggestion
    // The dependency closure of every module that transitively depends on this one: exactly
    // the compile contexts SwiftPM would have compiled these headers in.
```

This comment is not docc comment

Sources/ScipioKit/Resolver/PackageResolver.swift
```diff
@@ -107,9 +107,27 @@ actor PackageResolver {
...
+                metadata: .color(.yellow)
+            )
+            return nil
+        }
+
         return (allPackages: allPackages, allModules: allModules)
     }
 
+    /// Caches written before system-library support carry a wrong module type for system
+    /// modules; packaging from them would produce empty frameworks.
```
S-Shimotori: ditto
```suggestion
    // Caches written before system-library support carry a wrong module type for system
    // modules; packaging from them would produce empty frameworks.
```

Sources/ScipioKit/DescriptionPackage.swift
```diff
@@ -116,11 +128,14 @@ private final class BuildProductsResolver {
...
 
-        return try DependencyGraph<BuildProduct>.resolve(
+        // Resolve the full set first so every child ID stays resolvable; `filter` reconnects
+        // parents to children, preserving the build order of the remaining products.
+        let graph = try DependencyGraph<BuildProduct>.resolve(
             Set(products),
             id: \.target.name,
             childIDs: { $0.target.dependencies.flatMap(\.moduleNames) }
         )
+        return graph.filter { $0.target.underlying.type.isFrameworkProducible }
```
S-Shimotori: I guess this line is not covered by test

Sources/ScipioKit/Resolver/ModuleTypeResolver.swift
```diff
@@ -54,6 +59,32 @@ extension PackageResolver {
...
+            dependencyPackage: DependencyPackage
+        ) -> ResolvedModuleType {
+            assert(target.type == .system)
+
+            // SwiftPM expects module.modulemap at the module root and resolves its header paths
+            // against it, so the module directory itself is the include root.
+            let moduleFullPath = resolveTargetFullPath(of: target, dependencyPackage: dependencyPackage)
+                .standardizedFileURL
+            // Sorted because the enumeration order is unspecified: keeps the resolution
+            // result, and thus the serialized snapshot, deterministic.
```
S-Shimotori, line 73: How about adding a test to check this case?

Sources/ScipioKit/Producer/SystemLibraryPackager.swift
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    func createXCFramework(buildProduct: BuildProduct, outputDirectory: URL, overwrite: Bool) async throws {
+        let target = buildProduct.target
+
+        guard case let .system(includeDir, publicHeaders, moduleMapPath) = target.resolvedModuleType else {
+            throw Error.unexpectedModuleType(targetName: target.name)
+        }
+        guard fileSystem.exists(moduleMapPath) else {
+            throw Error.moduleMapNotFound(targetName: target.name, expectedPath: moduleMapPath)
```
S-Shimotori, line 59: Maybe we don't have a test for this line

Sources/ScipioKit/Producer/SystemLibraryPackager.swift
```diff
@@ -0,0 +1,255 @@
...
+            sourcePath,
+            string: "char _scipio_system_library_stub_\(target.c99name) = 0;\n"
+        )
+
+        var objectPaths: [URL] = []
+        for architecture in sdk.stubBinaryArchitectures {
+            let objectPath = workingDirectory.appending(component: "stub_\(architecture).o")
+            try await executor.execute([
+                "/usr/bin/xcrun",
+                "--sdk", sdk.settingValue,
```
S-Shimotori: I'm not familiar with mac catalyst, but I'm worried whether this command can handle mac catalyst or not

Sources/ScipioKit/Producer/PIF/FrameworkModuleMapGenerator.swift
```diff
@@ -176,12 +181,10 @@ struct FrameworkModuleMapGenerator {
...
         }
-        // TODO: Use modern regex
-        let regex = try NSRegularExpression(pattern: "^module", options: [])
-        let replaced = regex.stringByReplacingMatches(in: contents,
-                                                      range: NSRange(location: 0, length: contents.utf16.count),
-                                                      withTemplate: "framework module")
-        return replaced
+        // Line-anchored so declarations below leading comment lines convert too; nested
+        // submodules are indented and stay untouched.
+        let moduleDeclaration = /^module/.anchorsMatchLineEndings()
```
S-Shimotori: (nit) I prefer `/^module\b/`, `\s` or something

Sources/ScipioKit/Producer/SystemLibraryPackager.swift
Ckitakishi: Fixed in 960a9df.

Sources/ScipioKit/Resolver/PackageResolver.swift
Ckitakishi: Fixed in 960a9df, reworded to lead with what it detects.

Sources/ScipioKit/DescriptionPackage.swift
Ckitakishi: Covered in bf2a171. Writing it surfaced a semantics issue: a137657 prunes the whole subtree instead of reconnecting, and 661df21 aligns link references with it.

Sources/ScipioKit/Resolver/ModuleTypeResolver.swift
Ckitakishi: Covered in c1d3358: a nested header asserts recursive discovery in path-sorted order.

Sources/ScipioKit/Producer/SystemLibraryPackager.swift
Ckitakishi: Covered in 2c1530c, together with the unexpectedModuleType path.

Sources/ScipioKit/Producer/SystemLibraryPackager.swift
Ckitakishi: Good catch: `maccatalyst` is not an SDK name, so this failed under Catalyst. Fixed in 2c1530c: the stub compiles against the macOS SDK with a macabi target triple, covered by `packagesMacCatalystSlice`.

Sources/ScipioKit/Producer/PIF/FrameworkModuleMapGenerator.swift
Ckitakishi: Fixed in c752938.

S-Shimotori (approved):

